PR #1, opened 2012-08-16
Kimth/xqueue

kimth:
Let me know when these changes get deployed, then I'll update the pullonce example to hit the sandbox xqueue

### DIFF
--- a/queue/lms_interface.py
+++ b/queue/lms_interface.py
@@ -7,7 +7,6 @@ from boto.s3.connection import S3Connection
 from boto.s3.key import Key
 import json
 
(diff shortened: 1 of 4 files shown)


queue/lms_interface.py
```diff
@@ -7,6 +7,7 @@
 from boto.s3.key import Key
 import json
 
+from xqueue.aws_settings import AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY
```
cpennington: use `django.conf.settings`, rather than `xqueue.aws_settings`, otherwise this won't work in dev environments.

In particular, I think I'd leave the stuff alone down at the bottom, and just import `settings` at the top again

PR #4, opened 2012-08-26
Kimth/notify lms

kimth:
Xqueue management commands to perform the following:
- Requeue pulled submissions if it has been pulled and outstanding for some specific timeout (requested by Berkeley). TODO: requeue pushed submissions (6.00x).
- Retire submissions if the submission has reached a threshold number of failures. Queue sends message to the LMS/student that the submission will not be further processed, and a resubmit of the problem is needed.

No updates to the database model was made, so the above functionality can be retroactively applied. We will want to be careful with the requeueing of all submissions that already exist in the database, though.

The queueserver will need to run the management commands as cron jobs.


cpennington:
@ormsbee: Can you review this?

queue/consumer.py
```diff
@@ -48,6 +49,25 @@ def get_single_qitem(queue_name):
...
+        and that the problem should be resubmitted
+    '''
+    
+    # This is the only part of the XQueue that assumes knowledge of the external 
+    #   grader message format. TODO: Make the notification message-format agnostic
+    msg  = '<span>'
+    msg += 'Your submission could not be graded. '
+    msg += 'Please recheck your submission and try again. '
+    msg += 'If the problem persists, please notify the course staff.'
+    msg += '</span>'
```
ormsbee, line 64: Please push just the message and not the span tags formatting for it (so the LMS can display it however it wants).

kimth, line 64: The `<span>` was put there so that the message forms valid XML, which the LMS will test for. (I'm open to suggestions for other tags and/or other validation methods.) The grader can brick the problem state by providing unrenderable HTML (e.g. broken XML tags).

arjun810, line 64: I think this should keep the tags for now, and we can get rid of it once
invalid XML doesn't brick everything.

queue/management/commands/retire_submissions.py
```diff
@@ -0,0 +1,46 @@
...
+            action='store_true',
+            dest='force',
+            default=False,
+            help='Force retire submissions'),
+        )
+
+    def handle(self, *args, **options):
+        log.info(' [*] Scanning Submission database to retire failed submissions...')
+        
+        if len(args) > 1:
```
ormsbee: If having more than one arg does nothing, please explicitly error so that people don't think something's worked when it hasn't (e.g. if they put in two queue names).

queue/consumer.py
```diff
@@ -83,8 +103,8 @@ def _http_post(url, data):
         auth = None
 
     try:
-        r = requests.post(url, data=data, auth=auth)
-    except requests.exceptions.ConnectionError:
+        r = requests.post(url, data=data, auth=auth, timeout=settings.REQUESTS_TIMEOUT)
+    except (ConnectionError, Timeout):
```
ormsbee, line 107: Can we log what went wrong?

queue/management/commands/requeue_pulled_submissions.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class Command(BaseCommand):
+    args = "<queue_name>"
+    help = "Process currently pulled submissions, and requeue if external grader has not replied in settings.PULLED_SUBMISSION_TIMEOUT seconds. Optional <queue_name>" 
+
+    def handle(self, *args, **options):
+        log.info(' [*] Running requeue of pulled submissions...')
+
+        if len(args) > 0:
```
ormsbee: If having more than one arg does nothing, please explicitly error so that people don't think something's worked when it hasn't (e.g. if they put in two queue names).

queue/management/commands/requeue_pulled_submissions.py
```diff
@@ -0,0 +1,41 @@
...
+            queue_name = args[0]
+            open_submissions = Submission.objects.filter(queue_name=queue_name, lms_ack=False) 
+        else:
+            open_submissions = Submission.objects.filter(lms_ack=False)
+        open_submissions = open_submissions.exclude(pull_time=None)
+
+        for open_submission in open_submissions:
+            current_time = timezone.now()
```
ormsbee, line 33: In a lot of other places, we're using datetime.now(). I'm concerned about potentially mixing naive and timezone-aware dates. FWIW, our servers run on UTC, but we specify we're in America/New_York for the Django TIME_ZONE preference.

queue/consumer.py
ormsbee, line 64: Ah, got it, ty.

queue/management/commands/requeue_pulled_submissions.py
kimth, line 33: What's your recommendation here then?

queue/consumer.py
kimth, line 107: Done

queue/management/commands/retire_submissions.py
kimth: Changed so that the command processes list of queue names

queue/management/commands/requeue_pulled_submissions.py
kimth: Changed so that the command processes list of queue names

PR #7, opened 2012-09-22
Kimth/queue dos pull

kimth:
Queue-side DOS prevention. Idea:
- Submissions are tracked on a StudentModule (i.e. (user,module-id)) basis, which has a 1-to-1 mapping to the LMS callback URL. Note that we may want to strip the protocol and the base URL from the comparison URL, e.g. `http://www.edx.org/path/to/<user>/<id>/score_update` to `/path/to/<user>/<id>/score_update`.

This pull request requires a DB migration of the queue service: two new columns are added. 
- `lms_callback_url`, so that we can query on it directly (previously, it was in a json.dump field called `xqueue_header`) 
- `retired` Boolean flag that marks a submission as being "done." (I've been using the term "retired.") In the migration of existing data, the retired flag should be marked as `True` in order to avoid replay of prior submissions.

For the push interface (i.e. 6.00x), the functional changes:
- If you pool a large number of submissions from a single StudentModule, then only the latest submission will actually be sent to the external grader.
- If the external push fails, the LMS will immediately be informed of the failure.

For the pull interface (i.e. Berkeley), the functional changes:
- The API is unaffected, so Berkeley doesn't need to change anything on their side.
- The API call `get_submission` will only deliver unretired submissions to the external grader.

Internal to the queue service:
- The management commands `retire_submissions` and `requeue_pulled_submissions` make use of the new `retired` column rather than the previously overloaded `lms_ack`.

TODO:
- The API call `get_queuelen` does not distinguish between retired and unretired submissions.

### DIFF
--- a/queue/lms_interface.py
+++ b/queue/lms_interface.py
@@ -48,17 +48,19 @@ def submit(request):
                     s3_urls.update({filename: s3_url})
 
                 # Track the submission in the Submission database
-                submission = Submission(requester_id=get_request_ip(request),    
+                submission = Submission(requester_id=get_request_ip(request),
+                                        lms_callback_url=lms_callback_url,
                                         queue_name=queue_name,
                                         xqueue_header=xqueue_header,
                                         xqueue_body=xqueue_body,
                                         s3_urls=json.dumps(s3_urls),
                                         s3_keys=json.dumps(s3_keys))
                 submission.save()
+                print submission
 
                 qitem  = str(submission.id) # Submit the Submission pointer to queue
                 qcount = queue.producer.push_to_queue(queue_name, qitem)
-                
+
                 # For a successful submission, return the count of prior items
                 return HttpResponse(compose_reply(success=True, content="%d" % qcount))
         
--- a/queue/lms_interface.py
+++ b/queue/lms_interface.py
@@ -38,6 +38,10 @@ def submit(request):
             if queue_name not in settings.XQUEUES:
                 return HttpResponse(compose_reply(False, "Queue '%s' not found" % queue_name))
             else:
+                # Limit DOS attacks by invalidating prior submissions from the
+                #   same (user, module-id) pair
+                _invalidate_prior_submissions(lms_callback_url)
+
                 # Check for file uploads
                 s3_keys = dict() # For internal Xqueue use
                 s3_urls = dict() # For external grader use
@@ -63,7 +67,16 @@ def submit(request):
 
                 # For a successful submission, return the count of prior items
                 return HttpResponse(compose_reply(success=True, content="%d" % qcount))
-        
+
+
+def _invalidate_prior_submissions(lms_callback_url):
+    '''
+    Check the Submission DB to invalidate prior submissions from the same
+        (user, module-id). This function relies on the fact that lms_callback_url
+        takes the form: /path/to/callback/<user>/<id>/...
+    '''
+    prior_submissions = Submission.objects.filter(lms_callback_url=lms_callback_url, retired=False)
+    print len(prior_submissions)
 
 def _is_valid_request(xrequest):
     '''
--- a/queue/lms_interface.py
+++ b/queue/lms_interface.py
@@ -26,7 +26,7 @@ def submit(request):
         return HttpResponse(compose_reply(False, 'Queue requests should use HTTP POST'))
     else:
         # queue_name, xqueue_header, xqueue_body are all serialized
-        (request_is_valid, queue_name, xqueue_header, xqueue_body) = _is_valid_request(request.POST)
+        (request_is_valid, lms_callback_url, queue_name, xqueue_header, xqueue_body) = _is_valid_request(request.POST)
 
         if not request_is_valid:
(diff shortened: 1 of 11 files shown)


queue/lms_interface.py
```diff
@@ -58,10 +63,20 @@ def submit(request):
...
+
+
+def _invalidate_prior_submissions(lms_callback_url):
+    '''
+    Check the Submission DB to invalidate prior submissions from the same
+        (user, module-id). This function relies on the fact that lms_callback_url
+        takes the form: /path/to/callback/<user>/<id>/...
+    '''
+    prior_submissions = Submission.objects.filter(lms_callback_url=lms_callback_url, retired=False)
+    prior_submissions.update(retired=True)
```
shnayder, line 78: todo: add a warning comment that update() won't call any pre/post_save signals, or other django logic.  Presumably this doesn't matter for now, but it might in the future...

queue/models.py
```diff
@@ -9,12 +9,12 @@ class Submission(models.Model):
...
     Representation of submission request, including metadata information
     '''
-    
     # Submission 
-    requester_id  = models.CharField(max_length=CHARFIELD_LEN_SMALL) # ID of LMS
-    queue_name    = models.CharField(max_length=CHARFIELD_LEN_SMALL)
-    xqueue_header = models.CharField(max_length=CHARFIELD_LEN_LARGE)
-    xqueue_body   = models.TextField()
+    requester_id     = models.CharField(max_length=CHARFIELD_LEN_SMALL) # ID of LMS
+    lms_callback_url = models.CharField(max_length=CHARFIELD_LEN_SMALL)
```
shnayder, line 14: where's the migration file?  Did you forget to add it?

xqueue/settings.py
```diff
@@ -19,7 +19,8 @@
 
 REQUESTS_TIMEOUT = 1 # seconds
 
-XQUEUES = {'test-pull': None}
+XQUEUES = {'MITx-6.00x': 'http://ec2-23-22-89-106.compute-1.amazonaws.com',
```
shnayder, line 22: what machine is this?

queue/lms_interface.py
```diff
@@ -48,7 +52,8 @@ def submit(request):
                     s3_urls.update({filename: s3_url})
 
                 # Track the submission in the Submission database
-                submission = Submission(requester_id=get_request_ip(request),    
+                submission = Submission(requester_id=get_request_ip(request),
+                                        lms_callback_url=lms_callback_url,
```
ormsbee, line 56: When we migrate the existing DB, lms_callback_url for existing entries will be NULL (or some other default we can specify). Does this pull request handle that scenario?

xqueue/settings.py
kimth, line 22: This is my "local" pyxserver instance. Will remove.

queue/lms_interface.py
kimth, line 56: The queue will not break if the `lms_callback_url` column is NULL or invalid. This column is actually redundant, but has been extracted so that we can query directly on `lms_callback_url`, since it is 1-to-1 with a StudentModule in the LMS. (We may want a different name for the column.)

The redundancy is that `lms_callback_url` is part of the `xqueue_header` that was already written as a JSON dumps into the DB. Note that the actual callback will use the url from `xqueue_header` and not `lms_callback_url`.

On the other hand, there is a new Boolean flag called `retired` which marks submissions as being "done." In the migration, mark all existing submissions as `retired=True`.

kimth:
I'll verify the migration one more time tonight

ormsbee:
@kimth Any update on this?

kimth:
Hi Dave, sorry didn't have time to check this out. I'll ping you per email once re-checked.

queue/models.py
kimth, line 14: Added.

queue/migrations/0003_auto__add_field_submission_retired.py
```diff
@@ -0,0 +1,46 @@
...
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+        # Adding field 'Submission.retired'
+        db.add_column('queue_submission', 'retired',
+                      self.gf('django.db.models.fields.BooleanField')(default=False),
```
ormsbee: Shouldn't this default to True?

kimth: Here's the behavior that I want:
- The default value of the `retired` column for newly created Submissions should be `False`
- The default value of the `retired` column for existing Submissions that are being migrated should be `True`

ormsbee: Right. So that means your models.py file should specify the default is False (which it does). And I believe this line should specify default=True (since it describes the behavior for what happens to existing rows when they're migrated). keep_default should be left False as you have below.

kimth: Ok, I'm on this after working on CS169.1x and 6.002x...

PR #14, opened 2012-10-05
add indexes for commonly queried fields

ormsbee:


cpennington:
Doesn't this need an attached migration to do anything productive?

cpennington:
Or are you not storing the migrations in the code repo for xqueue?

ormsbee:
<cough> Er... yeah, just forgot to add the file. Yay code review. I've pushed it up.

PR #19, opened 2012-10-11
Remove the transaction from the consumer code.

shnayder:
- it should not be triggered now that we have proper behavior in the producer,
  and the transaction.commit_manually wrapper obscures exception logging

### DIFF
--- a/queue/consumer.py
+++ b/queue/consumer.py
@@ -192,10 +192,6 @@ class SingleChannel(threading.Thread):
         try:
             submission = Submission.objects.get(id=submission_id)
         except Submission.DoesNotExist:
-            # We'll complain about it soon...
-            pass
-
-        if submission == None:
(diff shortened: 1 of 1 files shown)


queue/consumer.py
```diff
@@ -179,29 +178,20 @@ def run(self):
...
+    # submission in the queue, this should be fine, but leaving here as a note.
+    # In particular, if we start seeing Submission.DoesNotExist errors and want
+    # to put back the "try again in 1 sec" logic, will need to manually manage
+    # transactions to make sure that retries see updated database state.
     def consumer_callback(self, ch, method, properties, qitem):
 
         submission_id = int(qitem)
 
         submission = None
-        for i in range(settings.DB_RETRIES):
```
kimth, line 192: Should also remove settings.DB_RETRIES

PR #21, opened 2012-10-24
Long s3 urls

kimth:
CS184.1x wants to upload up to 20 files at a time.

Allocating only 1024 characters to hold the serialized dict of s3 URLs is not enough when there are 20 files. So, convert the relevant fields to `TextField`

### DIFF
--- a/queue/migrations/0006_long_s3_urls.py
+++ b/queue/migrations/0006_long_s3_urls.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+import datetime
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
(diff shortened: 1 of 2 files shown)


ormsbee:
Altering table schemas can be very slow (it often requires rebuilding all the rows because the underlying format has changed), and we have > 1.4M rows in this table on prod right now. I'll look into it more tomorrow, but please don't merge/deploy this yet. I'm concerned we might end up locking the table for an hour. :-/

kimth:
I'll leave the merge/deploy to you. Meanwhile, I've suggested to cs184.1x that they should consider zipping up their >10 files before submission.

kimth:
Closing for time being. Need to increase `lms_callback` field length which we found previously is also being truncated (but without ill effect so far). Will be back up by evening

PR #23, opened 2012-12-04
Fix race condition that only happens in prod

ormsbee:
We thought it was safe to take out.

We thought wrong.

MANUAL TRANSACTIONS

Coming to a queuing server near you. Soon. Or already.


ormsbee:
ack, wrong branch to merge to

PR #24, opened 2012-12-04
Race condition hotfix

ormsbee:
We thought it was safe to take out.

We thought wrong.

MANUAL TRANSACTIONS

Coming to a queuing server near you. Soon. Or already.


ormsbee:
@shnayder Please review if you have a moment.

shnayder:
This looks fine, as far as it goes.  We still don't understand what causes this though, right?

PR #25, opened 2012-12-05
Use RABBIT_HOST in env.json if it exists

jarv:


shnayder:
code looks fine, though I'm not sure where this came from before, or the reasons for changing it now.

jarv:
Sorry should have mentioned in the pull request that this is for specifying a different host for the rabbitmq other than localhost which is the default in settings.  the intention is to run rabbitmq on a different set of servers.  The reason for doing this now is so that our rabbitmq servers are setup for high availability.

shnayder:
Ok, sounds good.  Merged.  Yay high availability.  Can we set up celery on them now?

jarv:
Yeah, I don't see why not.

PR #27, opened 2012-12-10
Feature/jarv/rabbituser

jarv:


shnayder:
Looks good to me

requirements.txt
```diff
@@ -8,3 +8,4 @@ boto
 path.py
 dogstatsd-python==0.2
 south
+MySQL-python
```
rocha, line 11: Didn't adding this requirement caused some problems in the past? (I am not sure, I overhead someone mentioning it).

jarv, line 11: I'm not sure, on the production queue servers we have MySQL-python==1.2.3
which was in the python requirements handled by puppet.
So far things look on the sandbox queue servers which are running in a
virtualenv and have this branch deployed to them.

shnayder, line 11: For mitx, this means that pip install doesn't "just work", and content
folks have to go download and install mysql (or more likely, get stuck and
need help).  For xqueue, that shouldn't be a problem.

jarv:
It looks like their aren't any issues so I am going to merge this.

PR #31, opened 2013-01-13
Retry LMS ack five times before giving up, in case of ELB disconnects.

ormsbee:
To help with push grader problems when ELB yanks LMS instances out of the pool abruptly.

### DIFF
--- a/queue/consumer.py
+++ b/queue/consumer.py
@@ -124,7 +124,12 @@ def post_grade_to_lms(header, body):
     lms_callback_url = header_dict['lms_callback_url']
 
     payload = {'xqueue_header': header, 'xqueue_body': body}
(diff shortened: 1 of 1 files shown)


queue/consumer.py
```diff
@@ -124,7 +124,12 @@ def post_grade_to_lms(header, body):
     lms_callback_url = header_dict['lms_callback_url']
 
     payload = {'xqueue_header': header, 'xqueue_body': body}
-    (success, lms_reply) = _http_post(lms_callback_url, payload, settings.REQUESTS_TIMEOUT)
+
+    attempts = 0
```
shnayder, line 132: Can you add a comment explaining why these are needed?  I'm unclear on why LMS machines going out of the ELB is a problem, as long as at least one is still there...  Is there a race condition somewhere?

PR #34, opened 2013-02-04
Feature/jarv/test push queues

jarv:


queue/management/commands/test_push.py
```diff
@@ -0,0 +1,71 @@
...
+logger = logging.getLogger(__name__)
+responses = {}
+
+
+def check_response(queue_name):
+    if queue_name in responses:
+        return True
+    else:
+        return False
+
```
rocha:
```
return queue_name in responses
```

queue/management/commands/test_push.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    def handle(self, *args, **options):
+        xq_client = XQueueClient(server='http://127.0.0.1:8000',
+                                 passwd=settings.XQUEUE_USERS['lms'],
+                                 post_url=
+                                 'http://stage-xqueue-001.m.edx.org:8989')
+        xq_client.login()
+        Hander = ServerHandler
+        httpd = TCPServerReuse(("", PORT), Hander)
```
rocha: Why do you need Handler as an alias,

```
httpd = TCPServerReuse(("", PORT), ServerHander)
```

rocha:
Some minor comments in the diff. 

A couple of questions:
- How are you detecting that there was no reply? Could it be improved by adding a timeout on the wait for the reply and logging the error?
- Somewhat related, Who is calling the command?

queue/management/commands/test_push.py
jarv: No reason, it started out as a copy and paste from the example in the docs http://docs.python.org/2/library/simplehttpserver.html
I'll change it.

jarv:
It's not timing out unfortunately if it doesn't hear anything back. I wasn't able to figure out a way to timeout the call to handle_request if there isn't any activity on the socket.  I tried to use threading.timer but that didn't work,  I can call a function after X number of seconds to check a watchdog but there doesn't seem to be a way to cleanly shutdown.

If you have any ideas for how to make it timeout after X number of seconds of inactivity that would be ideal.

The way this is run right now is as a Jenkins job http://jenkins.edx.org:8080/view/health-checks/job/stage-xqueue-test/ , the job fails by setting the build inactivity timeout which will force the abort.

rocha:
If find a recipe for TCPServerReuse that seems to do the timeout: http://code.activestate.com/recipes/425210/ 
It seems like it can be greatly simplified.

Another solution is to use signals to interrupt the process (kind of ugly)
I also found a recipe: http://code.activestate.com/recipes/534115-function-timeout/

Yet another alternative is to use tornado, less code but one more dependency.

jarv:
Yeah I saw the first link when I was looking into it. I think it's for
timing out connections once they have been made not timing out after not
seeing any connections. I tried using the code and it didn't seem to work.
I can raise an exception after X number of seconds but the process is still
hung.
I wasn't planning on spending a lot more time on this since it's doing what
I need it to do (continually test the queues from a jenkins job)

rocha:
You are right. I was just asking questions out of curiousity. I think it is good to merge.

PR #36, opened 2013-02-08
Allow consumers to pull from multiple queues, but take effort to make su...

cpennington:
...re a single queue can't overload all workers


queue/management/commands/run_consumer.py
```diff
@@ -1,13 +1,62 @@
...
+    if len(queue_assignments) < num_queues:
+        raise NoAssignmentError()
+
+    return dict(zip(queues, queue_assignments))
+
+
+def assign_queues_to_workers(num_workers, queue_assignments):
+    """
+    Takes a dictionary mapping queues to workers for those queues, and converts
+    it to a list of queues assigned to each worker
```
shnayder: converts to set, not list

cpennington:
Currently takes too long to start up, trying to figure out how to speed it up

queue/tests/test_run_consumer.py
```diff
@@ -0,0 +1,54 @@
...
+from itertools import combinations
+
+from queue.management.commands.run_consumer import assign_workers_to_queues, assign_queues_to_workers, NoAssignmentError
+
+
+def min_disjoint_workers(queue_assignments):
+    return min(len(set(a) - set(b)) for (a, b) in combinations(queue_assignments.values(), 2))
+
+
+def test_assign_workers_to_queues():
```
shnayder, line 11: Is there a jenkins job that actually runs these, btw?

cpennington, line 11: Yes. Just cleaned it up

queue/management/commands/run_consumer.py
cpennington: Fixed

PR #37, opened 2013-02-08
Cale/use multiprocessing

cpennington:


queue/management/commands/run_consumer.py
```diff
@@ -1,29 +1,97 @@
...
+            for (queue_name, push_url)
+            in settings.XQUEUES.items()
+            if push_url is not None
+        ]
+
+        workers_per_queue = settings.XQUEUE_WORKERS_PER_QUEUE
+        num_workers = settings.WORKER_COUNT
+
+        assignments = None
+        for min_disjoint_workers in range(workers_per_queue / 2, 0, -1):
```
ormsbee, line 79: If the settings for workers_per_queue = 1, does that mean that it'll never assign to anything (drop through this loop)?

PR #42, opened 2013-04-10
Feature/will/matlab integration tests

wedaly:
Integration tests for mathworks staging servers.
Basic integration test for graders that the XQueue pushes responses to.
Framework for developing additional integration tests, which we can later port to the celery implementation of xqueue.

Requires local env.json for Mathworks authentication information.
Also needs more work before it can run on Jenkins (unique queue names and database files in settings)

@rocha Please review when you get a chance.


xqueue/test_settings.py
```diff
@@ -10,19 +12,60 @@
...
+        # We also need to choose *unique* names to avoid
+        # conflicts in the Jenkins server
+        'TEST_NAME': 'test_xqueue_%s.sqlite' % uuid4().hex,
     }
 }
 
+# Local RabbitMQ configuration
+RABBITMQ_USER = 'guest'
+RABBITMQ_PASS = 'guest'
+RABBIT_HOST = 'localhost'
```
rocha: I think we can also get these values from test_env.json. Something like this:

``` python
RABBITMQ_USER = ENV_TOKENS.get('RABBITMQ_USER', 'guest')
```

rocha:
Looks good to me, except for the failing tests. Is it because we have not configured the message queue settings?

wedaly:
@rocha Finally got Jenkins set up correctly, so the tests are passing now.  Am I good to merge?

rocha:
Yes, go ahead.

PR #43, opened 2013-04-12
Feature/will/matlab load tests

wedaly:
- Created load test for Matlab servers using multi-mechanize and xqueue integration testing framework.
- Included README explaining how to set up load tests using the framework
- Removed tests/connect.py, which repeats tests that are now included in the automated test suite.

@dianakhuang Please review when you get a chance


dianakhuang:
One minor note while i'm going through this. You seem to have uploaded `load_tests/matlab/.DS_Store` by accident.

load_tests/matlab/config.cfg
```diff
@@ -0,0 +1,13 @@
...
+rampup = 0
+results_ts_interval = 10
+progress_bar = on
+console_logging = off
+xml_report = off
+
+
+[user_group-1]
+threads = 3
+script = submit_plot.py
```
dianakhuang, line 13: Where does this configuration get used?

wedaly, line 13: It's used by multi-mechanize: `multimech-run matlab`

dianakhuang, line 13: It might make sense to drop an extra line in the README to let people know that this is the config file for multi-mechanize in case they want to change some of these parameters.

dianakhuang:
Other than that, it looks good.

PR #45, opened 2013-04-16
Close db connections manually at the end of every worker thread

rocha:
Django is not consistently closing database connections automatically
at the end of each thread, causing the number of connections to go
beyond the database limit on certain high load situations.


cpennington:
Looks good. Merge when ready

PR #46, opened 2013-04-17
Reconnect automatically if the connection was lost unexpectedly

rocha:


queue/consumer.py
```diff
@@ -197,16 +197,35 @@ def __init__(self, queue_name, worker_url):
...
-                                               host=settings.RABBIT_HOST)
-
+        self.parameters = pika.ConnectionParameters(heartbeat_interval=5,
+                                                    credentials=credentials,
+                                                    host=settings.RABBIT_HOST)
         self.channel = None
-        self.connection = pika.SelectConnection(parameters, self.on_connected)
+        self.connection = self.connect()
+
+        self.connect()
```
cpennington: Why are you calling `self.connect()` twice?

rocha: fixed.

cpennington:
Looks good to me

PR #49, opened 2013-05-29
Fixed TypeError found in xqueue consumer logs.

wedaly:
This is a _partial_ fix for the issue causing the XQueue workers to crash.

1) This exception was logged at the time of the crashes on both XQueue prod servers:

Traceback (most recent call last):
  File "/usr/lib/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/opt/wwc/xqueue/queue/consumer.py", line 254, in run
    self.connection.ioloop.start()
  File "/opt/edx/src/pika/pika/adapters/select_connection.py", line 101, in start
    self.poller.start()
  File "/opt/edx/src/pika/pika/adapters/select_connection.py", line 385, in start
    self.poll()
  File "/opt/edx/src/pika/pika/adapters/select_connection.py", line 440, in poll
    self._handler(fileno, event, write_only=write_only)
  File "/opt/edx/src/pika/pika/adapters/base_connection.py", line 283, in _handle_events
    self._handle_read()
  File "/opt/edx/src/pika/pika/adapters/base_connection.py", line 304, in _handle_read
    return self._handle_disconnect()
  File "/opt/edx/src/pika/pika/adapters/base_connection.py", line 213, in _handle_disconnect
    self._on_connection_closed(None, True)
  File "/opt/edx/src/pika/pika/connection.py", line 1068, in _on_connection_closed
    self._on_disconnect(self.closing[0], self.closing[1])
  File "/opt/edx/src/pika/pika/connection.py", line 1166, in _on_disconnect
    self._process_connection_closed_callbacks(reply_code, reply_text)
  File "/opt/edx/src/pika/pika/connection.py", line 1195, in _process_connection_closed_callbacks
    reason_code, reason_text)
  File "/opt/edx/src/pika/pika/callback.py", line 61, in wrapper
    return function(_tuple(args), *_kwargs)
  File "/opt/edx/src/pika/pika/callback.py", line 92, in wrapper
    return function(_args, *_kwargs)
  File "/opt/edx/src/pika/pika/callback.py", line 232, in process
    callback(_args, *_keywords)
TypeError: on_connection_closed() takes exactly 2 arguments (4 given)

2) The `TypeError` was caused by a change in the pika library in the version we were using.  The changes in this commit prevent this error from occurring by updating the number of arguments accepted by the `on_connection_close()` callback.

3) Further investigation revealed that `on_connection_close()` is called when the connection to RabbitMQ is interrupted.  (Verified by running the test suite, stopping the RabbitMQ server, and observing exceptions before/after the above fix).

4) I do not know why the connection to RabbitMQ is being interrupted.  The worker should attempt to reconnect, so it might be able to recover once the `TypeError` is corrected.  If not, I have added additional logging to verify that the worker exited, both with and without an exception.  If the issue occurs again, we should have more information to diagnose it.

I do not know how to effectively test these failure conditions with the current implementation.  Pika's callback mechanism makes it very, very difficult to test effectively.  For this reason, I have not included tests verifying the fix.

Suggested reviewers: @dianakhuang @rocha

### DIFF
--- a/queue/consumer.py
+++ b/queue/consumer.py
@@ -13,6 +13,7 @@ from django.utils import timezone
 from django.db import connection as db_connection
 
 import pika
+from pika.exceptions import AMQPConnectionError
 from statsd import statsd
 from requests.exceptions import ConnectionError, Timeout
 
@@ -212,12 +213,12 @@ class Worker(multiprocessing.Process):
 
         self.connection.channel(self.on_channel_open)
 
-    def on_connection_closed(self, method_frame):
+    def on_connection_closed(self, connection, reply_code, reply_text):
         """Invoked when the connection is closed unexpectedly."""
 
         log.warning('Connection closed, trying to reconnect: ({0}) {1}'.format(
-            method_frame.method.reply_code,
-            method_frame.method.reply_text,
+            reply_code,
+            reply_text,
         ))
 
         # Try to reconnect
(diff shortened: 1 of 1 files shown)


rocha:
Looks good to me, thanks for fixing the problem and adding the extra log messages.

Do you know if there were other changes in the pika API?

dianakhuang:
:sunglasses:

wedaly:
@rocha I checked the other method signatures against http://pika.readthedocs.org/en/0.9.13/examples/asynchronous_consumer_example.html
and they all match.  The other callbacks get run by the test suite, so I'm less worried about them.

PR #54, opened 2013-06-27
Add retries to push_to_queue producer

rocha:
Don't fail immediately when trying connecting to the Broker. Retry a few times with some small pause.

### DIFF
--- a/queue/producer.py
+++ b/queue/producer.py
@@ -1,31 +1,56 @@
-import pika
+import time
+
 from django.conf import settings
+
+import pika
+
+
+MAX_RETRIES = 5
+RETRY_TIMEOUT = 0.5
 
 
 def push_to_queue(queue_name, qitem=None):
-    '''
+    """
     Publishes qitem (serialized data) to a specified queue.
     Returns the number of outstanding messages in specified queue
-    '''
+    """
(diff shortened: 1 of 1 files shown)


jarv:
looks good

rocha:
@cpennington can you do a sanity check? This is a quick hack to help @jarv with the certificate problems he is having.

queue/producer.py
```diff
@@ -1,31 +1,56 @@
...
+    retries = 0
+    while True:
+        try:
+            connection = pika.BlockingConnection(parameters)
+        except pika.exceptions.AMQPConnectionError as e:
+            retries += 1
+            if retries >= MAX_RETRIES:
+                raise e
+            time.sleep(RETRY_TIMEOUT)
+            continue
```
cpennington, line 38: Don't need to `continue` here, since it's at the end of the block anyway

cpennington:
For a quick hack, I'd have liked to see the hack unobscured by the (nice to have) whitespace changes. :+1: nonetheless.

queue/producer.py
rocha, line 38: I added it to make the end of the block explicit (the same thing as using the else in the try statement, it is not necessary).

PR #55, opened 2013-06-28
add ability to override timezone in env.json

jbau:
@rocha  simple change to allow env.json to change the timezone.


rocha:
LGTM

PR #57, opened 2013-07-10
Freeze to working versions for python-termstyle and rednose.

feanil:
Versions previously pinned to are no longer downloadable.


ormsbee:
:+1:

PR #58, opened 2013-07-29
Changed python-termstyle requirement to version 0.1.10

wedaly:
Since the developer apparently deleted version 0.1.9 from PyPi:
https://pypi.python.org/pypi/python-termstyle/0.1.9

Oddly, I was still able to get version 0.1.9 installed when using pip version 1.3.  To recreate, make sure you are using pip >= 1.4 in a fresh virtualenv.

Reviewer: @feanil


feanil:
Looks good.

PR #61, opened 2013-08-01
updated base url from v1 to v2.

kern3020:
From our conversation on the google group, I know this document is about v2. I'm pretty sure the base URL should be updated from v1 to v2.

### DIFF
--- a/docs/checkers_api.rst
+++ b/docs/checkers_api.rst
@@ -41,7 +41,7 @@ Basic Usage Example
 ::
 
     Request:
-    GET /checker/v1/queue/{queue-name} HTTP/1.1
+    GET /checker/v2/queue/{queue-name} HTTP/1.1
     Host: api.edx.org
     ...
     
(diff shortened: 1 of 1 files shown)


jtauber:
I'll check with the actual authors, but I think this is still best considered v1 of the external checker API.

This is like XBlock is to XModule. We wouldn't say the XBlock spec is v2 of XBlock. It's just our "v2" component architecture, but it's v1 of XBlock.

jtauber:
@rocha / @nedbat care to comment on this?

kern3020:
Hello James, 

I thought it should be updated because the spec starts out saying "This document describes the in-progress "v2" of the checker API". Then the URLs contains v1 which is not explained in the spec.

Sincerely,
John

For ease of reference
https://github.com/edx/xqueue/blob/jtauber/checker-docs/docs/checkers_api.rst

rocha:
Ah. I think the confusion comes from the fact that the current API is called "XQueue" and the one in progress is know as "External Checkers API". For the later, this would be version 1.  Does that make sense?

jtauber:
@kern3020 yeah, I deliberately put the v2 in quotes because I just meant to convey it was the next generation approach to external checkers. Sorry for the confusion. I'll change the docs.

PR #64, opened 2013-08-23
Upload submission to one bucket instead of one per queue

rocha:
The paramater S3_BUCKET must be set to the destination bucket


rocha:
@fephsun @e0d sanity check? Thanks!

rocha:
The parameter S3_BUCKET must be set in the env.json during deployment to AWS.

fephsun:
Looks good to me.

rocha:
@cpennington can you take a look? We ran out of buckets on S3.

cpennington:
Looks good. Are we going to migrate the existing uploads in the disparate buckets to the central bucket? Also, do the graders know now what bucket the files are in? Or do we just hand them a url?

rocha:
Yes, we are going to unify the existing buckets into one. Right now, we have around 30 submission buckets.

We pass a URL to the graders, which should be ok, unless they are keeping a record on the graders side. After talking to the relevant parties, it seems that none is doing so. We should however provide an easy way to map from the old locations to the new ones.

feanil:
looks good to me.

PR #65, opened 2013-09-04
config-prefix: Allow to specify SERVICE_VARIANT, used as config prefix

antoviaque:
Several services are currently using the same configuration filenames `/opt/wwc/env.json` & `/opt/wwc/auth.json` files. The configuration files would get overwritten when this is used by several different services cohabiting on a single instance. This patch allows to use the same convention as on the LMS/CMS (role edxapp on the configuration repository) to prefix the configuration files, with the `SERVICE_VARIANT` environment variable.

Cf https://github.com/edx/configuration/pull/234

### DIFF
--- a/queue/management/commands/update_users.py
+++ b/queue/management/commands/update_users.py
@@ -8,6 +8,7 @@ Ensure that the right users exist:
 """
 import json
 import logging
+import os
 
 from django.core.management.base import BaseCommand
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


feanil:
@rocha this change looks good to me, can you verify and merge.

rocha:
Look goods to me. 

@feanil
- The upstart scripts for xqueue and xqueue_consumer should change too, correct?
- How are we going to deploy this patch?

feanil:
By default this doesn't change how we run in production.  It should be the
same but this will allow single instance versions to work.

PR #66, opened 2013-09-19
allow ora (or other pull-grader) to repeat (amend) its reply

jbau:
@rocha 

This one line changes allows a pull grader to repeat a call to put_result (with potentially different result), which will be plumbed back through to the lms.

Doing this because we want to peer grading responses to be returned potentially incrementally, over several replies (each one a superset of the previous).

### DIFF
--- a/queue/ext_interface.py
+++ b/queue/ext_interface.py
@@ -136,7 +136,6 @@ def put_result(request):
                 return HttpResponse(compose_reply(False,'Incorrect key for submission'))
 
             submission.return_time = timezone.now()
-            submission.pullkey = ''
             submission.grader_reply = grader_reply
 
             # Deliver grading results to LMS


VikParuchuri:
@jbau We would also need to make some LMS-side changes to get the incremental feedback.  It is a difficult proposition on the LMS side, because a student could have reset/moved to next step by the time the incremental feedback is posted to the xmodule.

rocha:
Look good to me.

jbau:
@VikParuchuri This is all to handle the case in ORA where we're now potentially giving peer graders "finished" submissions.  I wasn't going to try to pop up notification if they've moved on when new score updates arrive.  I was just thinking the new score_update would save and then the student to could discover the new feedback on their own.  It'd be strictly better than limiting them to the previous feedbacks.

I guess the case where this may become bug is if they've already reset.  if that updated feedback does go through to the lms, will that become the state for the post-reset problem.  If so then it is a bug, and we shouldn't do it.

(I think the place to squelch it is in ORA, though)  This change seems fine to have in xqueue.

VikParuchuri:
@jbau Definitely don't see an issue with having it in xqueue.  Just wanted to ensure that you knew that this wouldn't make incremental feedback work -- that would need LMS-side code changes.

jbau:
Yup understood.

PR #69, opened 2013-10-31
E0d/upgrade boto

e0d:
@jtauber @rocha @jzoldak requirements updates. rake test seems happy although I'm not sure what the coverage is like.


rocha:
The coverage is pretty bad. Another thing that we can only really test on stage.

e0d:
OK, that's too bad, will have to test my branch on stage then.

PR #70, opened 2013-12-05
Monitor consumer workers and restart them if necessary

rocha:


rocha:
@feanil @ormsbee please review :smile:

queue/management/commands/run_consumer.py
```diff
@@ -75,10 +79,47 @@ def handle(self, *args, **options):
...
+        failed_workers = []
+
+        for worker in workers:
+            exitcode = worker.exitcode
+
+            if exitcode is None: # the process is running
+                continue
+            elif exitcode == 1:  # the process failed
+                failed_workers.append(worker)
+            else:  # the process has finished (0) or was interrupted (>0)
```
ormsbee: Did you mean >1?

ormsbee:
The code looks fine as far as I can tell, but don't we have system functionality (in upstart, supervisord, or whatever we're using on these boxes) that could more robustly monitor and restart workers? How did you test this?

feanil:
looks good to me, I'm also curious about how you tested it.

queue/management/commands/run_consumer.py
rocha: I think it is actually (<0). 

From the multiprocessing doc:

> exitcode: The child’s exit code. This will be None if the process has not yet terminated. A negative value -N indicates that the child was terminated by signal N.

rocha: Fixed

rocha:
@ormsbee You are correct, currently the parent process runs under upstart. There were two problems with the code:
1. If a child worker died, it was not restarted. The parent would only exit when all the children died.
2. If there was a connectivity problem, all the children will die, the parent will exit and upstart will restart it right away. The parent will exit right away again if the connectivity problem persisted, and upstart will give up after a certain number of retries fails over a short period of time. 

The solution was to monitor each child worker inside the parent process and restart it as necessary, and to introduce some delay before exiting.

rocha:
@feanil there is no automated test for the worker monitor. I am not sure how to do that, I probably should talk to @wedaly about that.

I tested manually by starting the consumer and submitting jobs. I had some code that made the consumer failed if it received certain message, and checked if the monitor will restart it. Finally I made sure that sending signals to each child worker will not cause it to restart, and if the parent process was signaled, it will exit and not hang.

ormsbee:
:+1:

rocha:
@feanil currently in `/etc/init/xqueue_consumer.conf`

```
respawn
respawn limit 3 30
```

Should we relax the limits to make it more tolerant?

feanil:
@rocha I talked about that to @jarv and @e0d and both felt that the we should not relax this and setup notifications to get notified when this stops.

rocha:
@feanil makes sense. Note that now it takes at least 10 seconds for the process to end if it cannot connect to RabbitMQ.

Good to merge?

feanil:
:+1:

PR #71, opened 2013-12-13
Add logging for submissions that take a long time to grade.

feanil:
Passing GRADING_TIMEOUT to the http request is deceptive in that the transaction does not timeout
after the given time.  The grading timeout as it is currently used in the http request will time
out if we don't begin to receive a response from the server in the given time but the response
could take a long time to finish.

@rocha @jarv

### DIFF
--- a/queue/consumer.py
+++ b/queue/consumer.py
@@ -360,8 +360,13 @@ class Worker(multiprocessing.Process):
         submission.push_time = timezone.now()
         start = time.time()
         (grading_success, grader_reply) = _http_post(self.worker_url, json.dumps(payload), settings.GRADING_TIMEOUT)
-        statsd.histogram('xqueue.consumer.consumer_callback.grading_time', time.time() - start,
+        grading_time = time.time() - start
+        statsd.histogram('xqueue.consumer.consumer_callback.grading_time', grading_time,
                          tags=['queue:{0}'.format(self.queue_name)])
+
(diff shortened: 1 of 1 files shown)


queue/consumer.py
```diff
@@ -360,9 +360,14 @@ def _deliver_submission(self, submission):
...
         (grading_success, grader_reply) = _http_post(self.worker_url, json.dumps(payload), settings.GRADING_TIMEOUT)
-        statsd.histogram('xqueue.consumer.consumer_callback.grading_time', time.time() - start,
+        grading_time = time.time() - start
+        statsd.histogram('xqueue.consumer.consumer_callback.grading_time', grading_time,
                          tags=['queue:{0}'.format(self.queue_name)])
 
+        if grading_time > settings.GRADING_TIMEOUT:
+            log.error("Grading time above {} for submission. grading_time: {}s body: {} files: {}".format(settings.GRADING_TIMEOUT,
+                      grading_time, submission.xqueue_body, submission.s3_urls))
+
```
rocha, line 370: Should something happen when the timeout occurs? Currently it only logs the error message.

feanil, line 370: @rocha Yea, currently we just want to get more info for which grading events that take a long time.  We've seen data in datadog that indicates that we occasionally get grading requests that take above 5 minutes so I think better logging would help in figure out what is going on.

rocha:
:+1:

PR #72, opened 2013-12-17
Force consumer to die when connection to RabbitMQ is closed.

ormsbee:
We are seeing instances where we detect a disconnect, try to
reconnect to RabbitMQ, and then fail to do so. The reconnection
code fails without error and the worker gracefully exits, bypassing
the monitor code that would restart it. This change causes us to
always throw an FailOnDisconnectError when we get a disconnect,
which causes the worker to die. The monitor will then restart it.

[LMS-1612]

@rocha, @e0d

### DIFF
--- a/queue/consumer.py
+++ b/queue/consumer.py
@@ -23,6 +23,20 @@ from queue.models import Submission
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


rocha:
:rocket:

e0d:
:+1:

PR #74, opened 2014-01-10
add rabbitmq virtual host parameterization

jbau:
@rocha 
Adding rabbitmq virtual-host configurability.  Defaults to '/' which should be the default vhost.  @wedaly I'm pretty sure this won't break your test setup, right?


rocha:
:+1:

wedaly:
:+1:

PR #80, opened 2014-04-17
Added an option for blocking get requests, possible when running with gevent

davestgermain:
@nedbat @e0d


nedbat:
@rocha any opinions on this?

rocha:
@nedbat @davestgermain looks good! I'll add a couple of code review comments, but all minor.

A few observations before merging and deploying:
- `pika` has a been a pain in the past. Newer versions pegged the CPU at 100% usage without any load. Other versions had a memory leak in the AMQP bindings. Updating the version sounds like a good idea if these are no longer a problems. I would advice to do a test deployment and observe the resources consumption before going ahead. 
- We don't have a lot of operational experience with `gevent`. Combined with the point above it creates some risk. @davestgermain what are you trying to achieve by using `gevent` in this case? If the benefit out-weights the risk, then it should be fine to go ahead.
- Coordinate carefully with devops. `Xqueue` is one of the most flaky parts of the platform. I suggest to also do a load test. Since there are not many courses during the summer, it may be a good time to do the upgrade.

queue/consumer.py
```diff
@@ -93,21 +94,47 @@ def get_single_qitem(queue_name):
...
-
-    # qitem is the item from the queue
-    method, header, qitem = channel.basic_get(queue=queue_name)
+    result = (False, '')
+    with contextlib.closing(connection):
+        channel = connection.channel()
+        channel.queue_declare(queue=queue_name, durable=True)
+
+
+        if blocking:
```
rocha, line 137: Maybe it would make it nicer to refactor the fork into something like:

```
if blocking:
   result = _initialize_blocking_consuming()
else:
   result = _initialize_non_blocking_consuming()
```

requirements.txt
```diff
@@ -1,18 +1,19 @@
...
-gunicorn==0.16.1
+dogstatsd-python==0.4.1
+gunicorn==18.0
 newrelic==2.4.0.4
 nose==1.2.1
 nosexcover==1.0.7
-path.py==2.4.1
--e git://github.com/pika/pika.git@a731347fc0#egg=pika
+path.py==5.1
+pika==0.9.13
```
rocha: Careful when upgrading `pika`, we had many problems with it in the past.

davestgermain:
@rocha The point of this PR is to reduce the number of requests to xqueue when polling graders (xqueue-watcher) are used. They can request `/xqueue/get_submission?block=true` and wait until an item is ready. Otherwise, there's extra overhead with opening and closing rabbitmq and http connections every second.

rocha:
@davestgermain sounds reasonable. If you haven't done so, I would really encourage you to test it in a production like environment before deploying. Keep an eye on resources consumption, like I mentioned earlier, we had problems with pika using all the CPU or leaking memory.

One more question. This should work in the presence of a load balancer, correct? By default requests timeout after one minute. In theory the graders would simply retry. Maybe we can adjust the timeout threshold? Check with devops if you think it is relevant.

sarina:
@nedbat @e0d this PR has been open for over a year, with no action since August and Dave no longer works for us. Should I close this PR?

PR #87, opened 2014-07-28
adding trailing / to avoid 301 from append slash

e0d:
@symbolist I think this also makes sense.


symbolist:
Yup! :+1:

e0d:
@davestgermain @jarv second review please.

jarv:
:+1:

PR #90, opened 2014-08-05
Assume the test is being run in the correct environment.

feanil:
In the case of our jenkins job it is now using shining panda to manage virtual envs.


feanil:
@jzoldak this should fix the brokenness more generally I think.

jzoldak:
:thumbsup:

PR #98, opened 2015-09-16
Upgrade to Django 1.8

muhammad-ammar:
TNL-2820

[certificates (see dashboard with honor user)](http://ma.sandbox.edx.org/)
[code response problem](http://ma.sandbox.edx.org/courses/course-v1:edX+DemoX+Demo_Course/courseware/d8a6192ade314473a78242dfeedfbf5b/0a5b8385851e478a859bd4590bbe109b/)


queue/consumer.py
```diff
@@ -20,6 +20,8 @@
 from queue.producer import get_queue_length
 from queue.models import Submission
 
+from util import commit_manually
```
nedbat: Is this import used? You added it, but didn't add any lines that use it.

muhammad-ammar: @nedbat there is no direct replacement for `commit_manually` in django 1.8, so i am trying different things. I will replace unused code once done with changes.

muhammad-ammar:
@wedaly @symbolist This is ready for review. Thanks!

queue/migrations/0004_preserve_unpolled_submissions.py
```diff
@@ -1,49 +0,0 @@
...
-    def forwards(self, orm):
-        """By default, when the migration is run, all old entries are marked as being retired.
-        However, entries that have not been pulled or pushed should not be marked this way -- 
-        they represent submissions from the LMS that have yet to be passed onto a grader, either
-        through the pull (Berkeley) or push (6.00x) mechanisms. We have to make sure these are 
-        unretired. But we're only going to check things from the last day, since anything older
-        than that is probably still retired but not properly marked because of a bug."""
-        oldest_allowed_unretired = datetime.datetime.now() - datetime.timedelta(days=1)
-        sql = "update queue_submission set retired=0 where pull_time is null and push_time is null " + \
-              "and arrival_time > '{0}'".format(oldest_allowed_unretired.strftime('%Y-%m-%d %H:%M:%S'))
```
muhammad-ammar, line 18: @ormsbee We are upgrading from south to django 1.8 migrations. I have a question, why we need this raw sql execution. This will run once only and after this migration has been applied this will never run again. What is the expected behavior? Do we still need this or should i remove it?

ormsbee, line 18: I don't remember this at all. But based no the docstring I wrote, it should be safe to just get rid of it (i.e. keep the migration here, but make it do nothing). It's a data migration written three years ago, intentionally scoped to just one day's worth of data. We would never roll back any deployed environment like that, and even if we did, the window is so small that it doesn't matter. New installs wouldn't have any data to migrate.

My guess is that we had some horrible bug and had to add this in a hurry to patch things for a course.

queue/models.py
```diff
@@ -17,7 +18,7 @@ class Submission(models.Model):
     '''
     # Submission 
     requester_id     = models.CharField(max_length=CHARFIELD_LEN_SMALL) # ID of LMS
-    lms_callback_url = models.CharField(max_length=CHARFIELD_LEN_SMALL, db_index=True)
+    lms_callback_url = models.CharField(max_length=CHARFIELD_LEN_MEDIUM, db_index=True)
```
wedaly: Why this change?

muhammad-ammar: During testing on sandbox, getting exception [at](https://github.com/edx/xqueue/blob/master/queue/lms_interface.py#L82) because url length is greater then `128`

wedaly: I'm concerned about changing this, because the application max length might not match the database constraint (created with the old migrations).  It sounds like this is an existing issue (probably specific to sandboxes, which have longer hostnames).  Could we create a ticket and do this later?

ormsbee: Ugh, I feel like I'm flashing back on this. The following is vague recollection, and you'd have to look into it to be sure, but... I think it silently truncates this, even in prod. If I remember correctly, what happens is that this field gets truncated, but the actual contents are redundantly written as part of the `xqueue_header`, and it's the value from `xqueue_header` that's used to actually do the post back. So the primary use of the `lms_callback_url` is for doing searches and freaking out anyone examining the database.

muhammad-ammar: This is due to the weirdness of django 1.4 which truncates the data if it exceeds the `max_length` limit but django 1.8 raises exception for this.

muhammad-ammar:
@feanil @nedbat @muzaffaryousaf @symbolist @doctoryes @macdiesel Please review. Thanks!

queue/consumer.py
```diff
@@ -316,26 +317,25 @@ def on_done():
...
-                    submission_id
-                ))
-
-            # if item has been retired, skip grading
-            if submission and not submission.retired:
-                self._deliver_submission(submission)
-
-            # close transaction
-            transaction.commit()
+            with transaction.atomic():
```
feanil, line 324: @muhammad-ammar will this roll back if there is an exception within the `with` clause?   I'm not familiar with the behavior of the `transaction.atomic()` when it comes to errors.

doctoryes, line 324: @feanil Yes - all the work within this block will be done within a DB savepoint. If an unhandled exception occurs in this block, then everything in the savepoint created by this block will be rolled-back/discarded. If no exception occurs, all writes in this savepoint will be committed to the outer savepoint or transaction.

queue/consumer.py
```diff
@@ -316,26 +317,25 @@ def on_done():
...
+                        submission_id
+                    ))
+
+                # if item has been retired, skip grading
+                if submission and not submission.retired:
+                    self._deliver_submission(submission)
+
         except Exception as e:
             # We need a wide catch here to correctly rollback the
             # transaction and acknowledge the message if something
```
doctoryes: Comment needs to be changed - not rolling back in this catch anymore.

doctoryes:
One nit - otherwise, :+1:

muhammad-ammar:
@doctoryes Updated the comment.

xqueue/settings.py
```diff
@@ -13,6 +13,8 @@
 DEBUG = False
 TEMPLATE_DEBUG = DEBUG
 
+ALLOWED_HOSTS = ['*']
```
macdiesel, line 16: @feanil will will want to change this for prod yes?

feanil, line 16: We'll want this to be configurable so that we can override it in prod yes.

xqueue/aws_settings.py
```diff
@@ -16,6 +16,8 @@
 S3_BUCKET = ENV_TOKENS.get('S3_BUCKET',S3_BUCKET)
 S3_PATH_PREFIX = ENV_TOKENS.get('S3_PATH_PREFIX',S3_PATH_PREFIX)
 
+ALLOWED_HOSTS = ENV_TOKENS.get('ALLOWED_HOSTS', ALLOWED_HOSTS)
```
muhammad-ammar, line 19: @feanil Configurable now.

feanil:
:+1:

macdiesel:
:+1:

PR #99, opened 2015-09-22
increase lms_callback_url field length

muhammad-ammar:
In django 1.4,  a model's field value gets truncated if it exceeds `max_length` limit but django 1.8 raises exception if we try to do this. We need to do this before [upgrading to django 1.8](https://github.com/edx/xqueue/pull/98)

### DIFF
--- a/queue/migrations/0006_auto__chg_field_submission_lms_callback_url.py
+++ b/queue/migrations/0006_auto__chg_field_submission_lms_callback_url.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+import datetime
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+
+        # Changing field 'Submission.lms_callback_url'
+        db.alter_column('queue_submission', 'lms_callback_url', self.gf('django.db.models.fields.CharField')(max_length=255))
+
+    def backwards(self, orm):
+
+        # Changing field 'Submission.lms_callback_url'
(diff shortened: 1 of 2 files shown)


muhammad-ammar:
@wedaly @symbolist Please have a look.

wedaly:
:+1: on the code.  For deploying this, do we have a sense of how long the migration will take for the number of submissions we have in the prod database?

FYI: @feanil

muhammad-ammar:
@cpennington Is there a limit on length of opaque keys as they are being used here in lms_callback_url? Do you think we can fix this issue in some other way? Let's say, don't save the value for `lms_callback_url` or truncate it in python?

cpennington:
Our database tables that store `CourseKey`s and `UsageKey`s in the LMS are capped at 255 characters.

feanil:
Seems like that means we need to increase the callback URL length. Maybe to 1020?  I'm still not sure how long migrations will take, this feels like something we should test.  There are about 4 million rows in the prod xqueue table it looks like.  An initial test might be to populate an xqueue table and then try to run the migration.  If it is pretty fast on a sandbox(under 10 seconds) then I wouldn't worry about it in production.  If it is significantly slower, then we should re-test on a DB that is closer to production in terms of performance.

muhammad-ammar:
If we try to increase the field length beyond `255` then we will get [Specified key was too long; max key length is 767 bytes](http://stackoverflow.com/questions/1814532/1071-specified-key-was-too-long-max-key-length-is-767-bytes). `lms_callback_url` field is only used for searching, for POSTing back to LMS `lms_callback_url value` in `xqueue_header` is used. So one thing we can do is to manually truncate this field and don't change the db schema.

PR #101, opened 2015-12-02
execute select_for_update inside atomic block

muhammad-ammar:
The issue is [select_for_update](https://github.com/edx/xqueue/blob/master/queue/ext_interface.py#L125) should be inside an atomic block. The issue is always there and why it wasn't captured earlier because select_for_update is not supported for sqlite and we are using sqlite on travis. In the case of sqlite Django will not raise an exception.
https://docs.djangoproject.com/en/1.8/ref/models/querysets/#select-for-update

I am feeling bad for sqlite, we should not use sqlite for testing.

### DIFF
--- a/queue/ext_interface.py
+++ b/queue/ext_interface.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.contrib.auth.decorators import login_required
+from django.db import transaction
 from django.http import HttpResponse
 from django.views.decorators.csrf import csrf_exempt
 from django.utils import timezone
@@ -122,7 +123,8 @@ def put_result(request):
             return HttpResponse(compose_reply(False, 'Incorrect reply format'))
         else:
             try:
-                submission = Submission.objects.select_for_update().get(id=submission_id)
+                with transaction.atomic():
+                    submission = Submission.objects.select_for_update().get(id=submission_id)
             except Submission.DoesNotExist:
                 log.error("Grader submission_id refers to nonexistent entry in Submission DB: grader: {0}, submission_id: {1}, submission_key: {2}, grader_reply: {3}".format(
                     get_request_ip(request), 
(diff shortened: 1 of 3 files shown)


muhammad-ammar:
@feanil @doctoryes @symbolist @nedbat @macdiesel @muzaffaryousaf ^^

symbolist:
@muhammad-ammar One question are not all views wrapped in transactions? Also was this the case before the Django upgrade?

doctoryes:
@symbolist @muhammad-ammar Yesterday, @feanil saw this same error in the production xqueue instance:

https://openedx.atlassian.net/browse/PLAT-932

We had forgotten to define `ATOMIC_REQUESTS` for all DB connections in xqueue, which meant that all views were _not_ being wrapped in an `atomic` transaction. He fixed this in this PR:

https://github.com/edx/configuration/pull/2531

However, that fix (when tested on stage) yielded another exception:

https://openedx.atlassian.net/browse/PLAT-933

The exception above shouldn't be happening since that view is wrapped in a `@transaction.non_atomic_requests` decorator. Yet - it's happening. My current theory is that the ordering of the decorators needs to be changed - the `transaction.non_atomic_requests` decorator is first everywhere else that it's used.

Anyway, I'm working on reproducing/fixing the PLAT-933 issue in the context of evaluating an xqueue security issue.

queue/ext_interface.py
```diff
@@ -122,7 +123,8 @@ def put_result(request):
             return HttpResponse(compose_reply(False, 'Incorrect reply format'))
         else:
             try:
-                submission = Submission.objects.select_for_update().get(id=submission_id)
+                with transaction.atomic():
+                    submission = Submission.objects.select_for_update().get(id=submission_id)
```
doctoryes, line 127: While this fixes the exception, the transaction doesn't last long enough. The `select_for_update` needs to persist to the actual updates below. The correct solution is wrapping this entire view in a `transaction.atomic`, which will be achieved by this PR:  https://github.com/edx/configuration/pull/2531

PR #102, opened 2015-12-02
Fix transactional code for xqueue.

doctoryes:
Test xqueue with `ATOMIC_REQUESTS` enabled, with MySQL instead of sqlite (due to difference with `select_for_update()`), and using `TransactionTestCase` to truly test the transactional code therein.

@feanil @macdiesel @muhammad-ammar @symbolist  Please review.


xqueue/test_settings.py
```diff
@@ -19,16 +19,19 @@
 
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': 'test_xqueue.sqlite',
+        'ENGINE': 'django.db.backends.mysql',
```
macdiesel, line 22: Why the database backend change for testing?

doctoryes, line 22: @muhammad-ammar explains it in this PR: https://github.com/edx/xqueue/pull/101

feanil:
:+1:

queue/lms_interface.py
```diff
@@ -20,10 +20,11 @@
 
 log = logging.getLogger(__name__)
 
+
+@transaction.non_atomic_requests
```
macdiesel, line 24: does non_atomic_requests always need to be the outermost decorator?

macdiesel:
good stuff dude! Here, have a :goat: 

:+1:

queue/lms_interface.py
doctoryes, line 24: There's nothing that I've read that states that `non_atomic_requests` must be first. However, a) it's the only change I made that would affect the error that was fixed here - and b) all other places where the decorator is used (and those places appear to work) has it placed first among other decorators. It could be that `@statsd.timed` doesn't play nice with anything under it? Dunno - but I'd say put the `@transaction` decorators first as a rule.

symbolist, line 24: @macdiesel @doctoryes It is not documented but if you look at the code, it is clear it would only work if it is the outermost decorator: https://github.com/django/django/blob/master/django/core/handlers/base.py#L77

doctoryes, line 24: @symbolist I've looked at the mechanism for `non_atomic_requests` and it's still not clear to me. Wouldn't the set of `view._non_atomic_requests` still have the default DB in it at the time that `BaseHandler.get_response` is called? I assumed that it would be called after all the decorator wrapping, so it wouldn't matter if the decorator was first or last among other decorators.

symbolist, line 24: @doctoryes My understanding is that since some decorators return a new function object, if the decorator `non_atomic_requests` is not on the outside the attribute `_non_atomic_requests = True` would exist on an encapsulated function object instead of the one that is assigned to the module attribute (`submit` in this case).

PR #104, opened 2016-01-27
Make put_result atomic in order to use select_for_update in django 1.6+

defance:
Since django version was bumped to 1.8 from 1.4 some bugs still to be catched.

As I updated my instance of xqueue put_result stopped working. And the exception I got was related to `Submission.select_for_update()`. 

Since django 1.6 the behaviour was changed and from then on `select_for_update` can be used only within transaction. As a solution a block containing submission select and update can be marked atomic. This commits wraps the whole put_result handler atomic.

For more information see: http://stackoverflow.com/questions/17149587/select-for-update-in-development-django

### DIFF
--- a/queue/ext_interface.py
+++ b/queue/ext_interface.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.contrib.auth.decorators import login_required
+from django.db import transaction
 from django.http import HttpResponse
 from django.views.decorators.csrf import csrf_exempt
 from django.utils import timezone
@@ -109,6 +110,7 @@ def get_submission(request):
             return HttpResponse(compose_reply(True, content=json.dumps(payload)))
 
 
+@transaction.atomic
 @csrf_exempt
 @login_required
 @statsd.timed('xqueue.ext_interface.put_result.time')


openedx-webhooks:
Thanks for the pull request, @defance! I've created [OSPR-1093](https://openedx.atlassian.net/browse/OSPR-1093) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

If you like, you can add yourself to the [AUTHORS](https://github.com/edx/xqueue/blob/master/AUTHORS)      file for this repo, though that isn't required.     Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.

nedbat:
@doctoryes Can you take a look at this?

doctoryes:
It's indeed true that `select_for_update` needs to happen inside a transaction to be effective. However, the not-well-stated assumption for xqueue is that it will be running on a database where `ATOMIC_REQUESTS` is True.

You can see that in the xqueue test settings here:
https://github.com/edx/xqueue/blob/master/xqueue/test_settings.py#L36
and in the edx.org production settings here:
https://github.com/edx/configuration/blob/master/playbooks/roles/xqueue/defaults/main.yml#L82

Thinking through this, it seems that we shouldn't assume that `ATOMIC_REQUESTS` are turned on (though I'd advise it for your own instance). Adding this `transaction.atomic` makes the assumption more explicit, which is a good thing. So I'm in favor of this change - except for the location of the decorator (I'll add a separate comment for that).

queue/ext_interface.py
```diff
@@ -112,6 +113,7 @@ def get_submission(request):
 @csrf_exempt
 @login_required
 @statsd.timed('xqueue.ext_interface.put_result.time')
+@transaction.atomic
```
doctoryes: Please move this decorator to be the first decorator for the method. The location it's in now would work, but it's a standard I'd like to adopt. If `ATOMIC_REQUESTS` is True and you add a `transaction.non_atomic_requests` to turn off the transaction, it won't work - see: https://github.com/edx/xqueue/pull/102#discussion-diff-46598630

defance:
I moved atomic-decorator to be first in the list as it everywhere else.

Looks like my configuration repo was too old that I discovered this "atomic"-issue. Still I agree that one cannot rely on single configuration setting as its absence leads xqueue to non-working state.

doctoryes:
Thanks, @defance - could you please squash the two commits into one? Once that's done, this change is ready to merge.

defance:
@doctoryes Here you go. :)

doctoryes:
:+1: 

Thanks again, @defance !

PR #107, opened 2016-05-26
Add retries into consumer connection.

attiyaIshaque:
## [TNL-1434](https://openedx.atlassian.net/browse/TNL-1434)

Avoiding `pika.exceptions:AMQPConnectionError: (0, 'Not specified')` by adding retries for creating the connection. 
### Reviewers
- [x] @adampalay 
- [x] @macdiesel

### DIFF
--- a/queue/consumer.py
+++ b/queue/consumer.py
@@ -207,6 +207,7 @@ class Worker(multiprocessing.Process):
         # not get shared with the children threads.
         db_connection.close()
 
+        self.retries = 0
         self.id = next(self.counter)
         self.queue_name = queue_name
         self.worker_url = worker_url
@@ -229,6 +230,9 @@ class Worker(multiprocessing.Process):
         self.connection.add_on_close_callback(self.on_connection_closed)
 
         self.connection.channel(self.on_channel_open)
+
+        # Set the retires attempts to Zero
+        self.retries = 0
 
     def on_connection_closed(self, connection, reply_code, reply_text):
         """Invoked when the connection is closed unexpectedly."""
@@ -271,21 +275,44 @@ class Worker(multiprocessing.Process):
             queue=self.queue_name,
         ))
 
-        try:
-            self.connection = self.connect()
-            self.connection.ioloop.start()
-        except AMQPConnectionError as ex:
-            log.error("[{id}] Consumer for queue {queue} connection error: {err}".format(
-                id=self.id, queue=self.queue_name, err=ex))
-            raise
-        else:
-            # Log that the worker exited without an exception
-            log.info(" [{id}] Consumer for queue {queue} is exiting normally...".format(
-                id=self.id, queue=self.queue_name))
-        finally:
-            # Log that the worker stopped
-            log.info(" [{id}] Consumer for queue {queue} stopped".format(
-                id=self.id, queue=self.queue_name))
+        while True:
+            try:
+                log.info("[{id}] - Attempting to establish a connection for queue {queue}".format(
+                    id=self.id,
+                    queue=self.queue_name,
+                ))
+                self.connection = self.connect()
+                self.connection.ioloop.start()
+            except AMQPConnectionError as ex:
+                self.retries += 1
+                if self.retries >= settings.RETRY_MAX_ATTEMPTS:
+                    log.error("[{id}] Consumer for queue {queue} connection error: {err}".format(
+                        id=self.id,
+                        queue=self.queue_name,
+                        err=ex
+                    ))
+                    raise
+                log.info("[{id}] - Retrying connection, attempt # {attempt} of {max_attempts} of MAX".format(
+                    id=self.id,
+                    attempt=self.retries,
+                    max_attempts=settings.RETRY_MAX_ATTEMPTS
+                ))
+                if self.retries > 1:
+                    time.sleep(settings.RETRY_TIMEOUT)
+                continue
+            else:
+                # Log that the worker exited without an exception
+                log.info(" [{id}] Consumer for queue {queue} is exiting normally...".format(
+                    id=self.id,
+                    queue=self.queue_name
+                ))
+                break
+            finally:
+                # Log that the worker stopped
+                log.info(" [{id}] Consumer for queue {queue} stopped".format(
+                    id=self.id,
+                    queue=self.queue_name
+                ))
 
         # TODO [rocha] make to to finish all  submissions before exiting
 
--- a/queue/tests/test_consumer.py
+++ b/queue/tests/test_consumer.py
@@ -1,0 +1,48 @@
+import mock
+from unittest import TestCase
+from django.conf import settings
+from uuid import uuid4
+
+from pika.exceptions import AMQPConnectionError
+
+from queue.consumer import Worker
+
+
+class TestWorkerConnection(TestCase):
+
+    QUEUE_NAME = 'test_queue_%s' % uuid4().hex
+
+    def setUp(self):
+        self.worker = Worker(queue_name=self.QUEUE_NAME, worker_url='some_test_url')
+
+    @mock.patch('time.sleep', mock.Mock(return_value=None))
+    @mock.patch('queue.consumer.log')
+    @mock.patch('queue.consumer.Worker.connect')
+    def test_connection_retries_on_exception(self, mock_worker_connect, mock_log):
+        """
+        Tests worker's connection on 'AMQPConnectionError' exception.
+        """
+        mock_worker_connect.side_effect = AMQPConnectionError
+
+        self.assertEquals(self.worker.retries, 0)
+
+        with self.assertRaises(AMQPConnectionError):
+            self.worker.run()
+
+        self.assertEquals(self.worker.retries, settings.RETRY_MAX_ATTEMPTS)
+
+        # Asserts connection retry logging.
+        for attempt in xrange(1, settings.RETRY_MAX_ATTEMPTS):
+            mock_log.info.assert_any_call(
+                "[{id}] - Retrying connection, attempt # {attempt} of {max_attempts} of MAX".format(
+                    id=self.worker.id,
+                    attempt=attempt,
+                    max_attempts=settings.RETRY_MAX_ATTEMPTS
+                )
+            )
+
+        # Asserts that the error was logged on crossing max retry attempts.
+        mock_log.error.assert_called_with("[{id}] Consumer for queue {queue} connection error: ".format(
+            id=self.worker.id,
+            queue=self.worker.queue_name,
+        ))
--- a/xqueue/aws_settings.py
+++ b/xqueue/aws_settings.py
@@ -1,5 +1,6 @@
 from settings import *
 import json
+import os
 from logsettings import get_logger_config
 
 # Allow to specify a prefix for env/auth configuration files
@@ -42,3 +43,6 @@ REQUESTS_BASIC_AUTH = AUTH_TOKENS["REQUESTS_BASIC_AUTH"]
 RABBITMQ_USER = AUTH_TOKENS.get('RABBITMQ_USER', 'guest').encode('ascii')
 RABBITMQ_PASS = AUTH_TOKENS.get('RABBITMQ_PASS', 'guest').encode('ascii')
 XQUEUE_USERS = AUTH_TOKENS.get('USERS', None)
+
+RETRY_MAX_ATTEMPTS = os.environ.get('RETRY_MAX_ATTEMPTS', 10)
+RETRY_TIMEOUT = os.environ.get('RETRY_TIMEOUT', 10)
--- a/xqueue/aws_settings.py
+++ b/xqueue/aws_settings.py
@@ -1,6 +1,5 @@
 from settings import *
 import json
-import os
 from logsettings import get_logger_config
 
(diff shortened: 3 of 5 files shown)


queue/consumer.py
```diff
@@ -271,21 +273,32 @@ def run(self):
...
+                        queue=self.queue_name, err=ex
+                    ))
+                    raise
+                time.sleep(RETRY_TIMEOUT)
+                continue
+            else:
+                # Log that the worker exited without an exception
+                log.info(" [{id}] Consumer for queue {queue} is exiting normally...".format(
+                    id=self.id, queue=self.queue_name
+                    ))
```
adampalay: nit, this should be shifted left

queue/consumer.py
```diff
@@ -271,21 +273,32 @@ def run(self):
...
+                break
+            finally:
+                # Log that the worker stopped
+                log.info(" [{id}] Consumer for queue {queue} stopped".format(
+                    id=self.id, queue=self.queue_name
+                    ))
```
adampalay: nit, this should be shifted left

adampalay:
👍 , this should have an ops review too!

attiyaIshaque:
@maxrothman ???

queue/consumer.py
```diff
@@ -23,6 +23,8 @@
 
 log = logging.getLogger(__name__)
 
+MAX_RETRIES = 5
+RETRY_TIMEOUT = 0.5
```
macdiesel: Why such a short pause between retries?

queue/consumer.py
```diff
@@ -271,21 +273,32 @@ def run(self):
...
-            log.info(" [{id}] Consumer for queue {queue} stopped".format(
-                id=self.id, queue=self.queue_name))
+        retries = 0
+        while True:
+            try:
+                self.connection = self.connect()
+                self.connection.ioloop.start()
+            except AMQPConnectionError as ex:
+                retries += 1
+                if retries >= MAX_RETRIES:
```
macdiesel: The thing that stands out to me here is that after a successful reconnection the retries count does not reset. what could happen is that you hit hit some error condition one day, the retry count gets up to 5 but reconnects on the last attempt. Then some amount of time passes, say a week, and this code never retries the connection since the ceiling was met a few days prior.

queue/consumer.py
macdiesel: Please make these variables settable as environment variables.

queue/consumer.py
adampalay: @macdiesel there's a `break` on L296. So the next time that `run` is called, the retry count will reset. It may make more sense to write this as a for loop to avoid this kind of confusion.

queue/consumer.py
```diff
@@ -271,21 +273,32 @@ def run(self):
...
-            # Log that the worker stopped
-            log.info(" [{id}] Consumer for queue {queue} stopped".format(
-                id=self.id, queue=self.queue_name))
+        retries = 0
+        while True:
+            try:
+                self.connection = self.connect()
+                self.connection.ioloop.start()
+            except AMQPConnectionError as ex:
+                retries += 1
```
adampalay: @attiyaIshaque , I think it'll be worth logging when a try happens

queue/consumer.py
adampalay: @macdiesel that's an interesting point! Sounds like we should make sure the retries reset after some time window, like a few seconds.

macdiesel:
some other things to think about here:
- What happens when  def stop(self): is called? it will fire the on_connection_closed method. There should be something in here to ensure that the retry is not restarted when stop is called. Should be able to handle this with a boolean.
- the retry attempts can be reset in the on_connected method. 
- I would increase both the number of attempts as well as the delay between attempts. Perhaps 5 or more seconds between retries. We don't want to DDOS ourselves.
- More logging around **everything** involved in retrying the connection.

queue/consumer.py
attiyaIshaque: @macdiesel Do you mean adding the variables in _.env.json_ file? If that's so, who will add those configuration variables to the file?
FYI: @adampalay

attiyaIshaque:
@macdiesel I am going to add reset attempt code in  `on_connected` method. Thank you for the recommendations.

queue/consumer.py
macdiesel: import os
os.environ.get('RETRY_MAX_ATTEMPTS', 15)

attiyaIshaque:
@adampalay and @macdiesel I have addressed the comments.

queue/consumer.py
```diff
@@ -197,6 +198,7 @@ class Worker(multiprocessing.Process):
     """Encapsulation of a single RabbitMQ listener that listens on a queue
     """
     counter = itertools.count()
+    retires = 0
```
macdiesel: Please move this inside the `__init__` method.

xqueue/aws_settings.py
```diff
@@ -42,3 +43,6 @@
 RABBITMQ_USER = AUTH_TOKENS.get('RABBITMQ_USER', 'guest').encode('ascii')
 RABBITMQ_PASS = AUTH_TOKENS.get('RABBITMQ_PASS', 'guest').encode('ascii')
 XQUEUE_USERS = AUTH_TOKENS.get('USERS', None)
+
+RETRY_MAX_ATTEMPTS = os.environ.get('RETRY_MAX_ATTEMPTS', 5)
+RETRY_TIMEOUT = os.environ.get('RETRY_TIMEOUT', 0.5)
```
macdiesel: Please set this to 10 seconds.

xqueue/aws_settings.py
```diff
@@ -42,3 +43,6 @@
 RABBITMQ_USER = AUTH_TOKENS.get('RABBITMQ_USER', 'guest').encode('ascii')
 RABBITMQ_PASS = AUTH_TOKENS.get('RABBITMQ_PASS', 'guest').encode('ascii')
 XQUEUE_USERS = AUTH_TOKENS.get('USERS', None)
+
+RETRY_MAX_ATTEMPTS = os.environ.get('RETRY_MAX_ATTEMPTS', 5)
```
macdiesel: Please set this to 10.

queue/consumer.py
```diff
@@ -271,21 +276,35 @@ def run(self):
...
-            # Log that the worker exited without an exception
-            log.info(" [{id}] Consumer for queue {queue} is exiting normally...".format(
-                id=self.id, queue=self.queue_name))
-        finally:
-            # Log that the worker stopped
-            log.info(" [{id}] Consumer for queue {queue} stopped".format(
-                id=self.id, queue=self.queue_name))
+        while True:
+            try:
+                log.info(" [{id}] Establishing connection consumer for queue {queue}".format(
```
macdiesel: change to: "[{id}] - Attempting to establish a connection for queue {queue}"

queue/consumer.py
```diff
@@ -271,21 +276,35 @@ def run(self):
...
+                    id=self.id,
+                    queue=self.queue_name,
+                ))
+                self.connection = self.connect()
+                self.connection.ioloop.start()
+            except AMQPConnectionError as ex:
+                self.retries += 1
+                if self.retries >= settings.RETRY_MAX_ATTEMPTS:
+                    log.error("[{id}] Consumer for queue {queue} connection error: {err}".format(
+                        id=self.id,
```
macdiesel: Please add a logging statement stating that the connection will be retried and the current retry attempt. The retry count should be attempting retry n RETRY_MAX_ATTEMPTS.

queue/consumer.py
```diff
@@ -271,21 +276,35 @@ def run(self):
...
+                        queue=self.queue_name, err=ex
+                    ))
+                    raise
+                time.sleep(settings.RETRY_TIMEOUT)
```
macdiesel: Please add a logging statement that states the retry attempt number and what RETRY_MAX_ATTEMPTS is set to. "Retrying connection, attempt n of MAX."

macdiesel: Also the first retry attempt should not sleep.

queue/consumer.py
```diff
@@ -271,21 +276,35 @@ def run(self):
...
+                    queue=self.queue_name,
+                ))
+                self.connection = self.connect()
+                self.connection.ioloop.start()
+            except AMQPConnectionError as ex:
+                self.retries += 1
+                if self.retries >= settings.RETRY_MAX_ATTEMPTS:
+                    log.error("[{id}] Consumer for queue {queue} connection error: {err}".format(
+                        id=self.id,
+                        queue=self.queue_name, err=ex
```
macdiesel: err should be on the next line.

macdiesel:
@attiyaIshaque I have finished this pass of my reivew. Thanks!

queue/consumer.py
adampalay: this is also a typo — should be `self.retries`. I'm starting to think this is a big enough change that it would be wise to add tests.

attiyaIshaque:
@macdiesel I have addressed the comments.

queue/consumer.py
```diff
@@ -271,21 +276,40 @@ def run(self):
...
+            except AMQPConnectionError as ex:
+                self.retries += 1
+                if self.retries >= settings.RETRY_MAX_ATTEMPTS:
+                    log.error("[{id}] Consumer for queue {queue} connection error: {err}".format(
+                        id=self.id,
+                        queue=self.queue_name,
+                        err=ex
+                    ))
+                    raise
+                log.info("Retrying connection, attempt {max_attempt} of MAX".formar(
```
macdiesel: Typo, should be s/formar/format. 
Let me just give you the line that it should be:
`log.info("[{id}] - Retrying connection, attempt # {attempt} of {max_attempts} of MAX".format(id=self.id, attempt=self.retries, max_attempts=settings. RETRY_MAX_ATTEMPTS))`

queue/consumer.py
```diff
@@ -7,6 +7,7 @@
 import time
 import itertools
 
+
```
macdiesel: remove this line.

macdiesel:
Reviewed again. I agree with @adampalay tests would be good. Found another typo this round.

@attiyaIshaque if you do not already you should consider using an IDE. It would help you clear up these simple mistakes. 
https://www.jetbrains.com/pycharm/download/

Community edition is free and will at least get you the simple syntax checking and code highlighting.

attiyaIshaque:
@macdiesel I am already using IDE but it very strange why it does not highlight this `formar(` syntax error.

attiyaIshaque:
@macdiesel I have addressed the comment.

Qubad786:
@macdiesel , @adampalay may you guys please review again, I've added test this time. Thanks!

xqueue/aws_settings.py
```diff
@@ -1,5 +1,6 @@
 from settings import *
 import json
+import os
```
Qubad786: ouch! – forgot to remove this redundant import.

Qubad786:
@adampalay, @macdiesel it is ready for review.

xqueue/aws_settings.py
Qubad786: I'll remove this at-once in addressing of yet to come review.

queue/consumer.py
```diff
@@ -271,21 +275,44 @@ def run(self):
...
-            raise
-        else:
-            # Log that the worker exited without an exception
-            log.info(" [{id}] Consumer for queue {queue} is exiting normally...".format(
-                id=self.id, queue=self.queue_name))
-        finally:
-            # Log that the worker stopped
-            log.info(" [{id}] Consumer for queue {queue} stopped".format(
-                id=self.id, queue=self.queue_name))
+        while True:
```
adampalay, line 278: @Qubad786 , what do you think of doing a `for` loop instead of a `while` loop? It may be more readable.

adampalay:
@Qubad786 , I made a small comment/suggestion, but this is good to go as is if you'd like! 👍

Qubad786:
Hey @macdiesel , will you be able to review this? Waiting on your response. Thanks!

queue/tests/test_consumer.py
```diff
@@ -0,0 +1,47 @@
...
+
+class TestWorkerConnection(TestCase):
+
+    QUEUE_NAME = 'test_queue_%s' % uuid4().hex
+
+    def setUp(self):
+        self.worker = Worker(queue_name=self.QUEUE_NAME, worker_url='some_test_url')
+
+    @mock.patch('queue.consumer.log')
+    @mock.patch('queue.consumer.Worker.connect')
```
macdiesel, line 20: Lets also patch time.sleep and have it return immediately so the test will run faster.

macdiesel:
One comment to address. Great everywhere else. Thanks @Qubad786  👍

queue/tests/test_consumer.py
macdiesel, line 20: or set RETRY_TIMEOUT to 0. Either way.

Qubad786, line 20: Great point! I missed it.

queue/consumer.py
Qubad786, line 278: I feel like `while` is good fit in this particular case.

Qubad786:
@macdiesel I've addressed your comment. I will squash and merge now. Thanks!

PR #130, opened 2018-01-05
Make logging consistent with platform

coryleeio:
Our builds are failing atm. Making the logging configuration consistent with edxapp fixes the issue.

### DIFF
--- a/xqueue/logsettings.py
+++ b/xqueue/logsettings.py
@@ -62,12 +62,6 @@ def get_logger_config(log_dir,
                 'formatter': 'standard',
                 'stream': sys.stdout,
             },
-            'syslogger-remote': {
-                'level': 'INFO',
-                'class': 'logging.handlers.SysLogHandler',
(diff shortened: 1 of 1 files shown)


xqueue/logsettings.py
```diff
@@ -41,6 +41,8 @@ def get_logger_config(log_dir,
                         logging_env=logging_env, hostname=hostname)
 
     handlers = ['console', 'local'] if debug else ['local']
+    if syslog_addr:
+        handlers.append('syslogger-remote')
```
nasthagiri, line 45: Are these lines still needed?

coryleeio: Handlers is - it's used on lines 68 and 73 - As for conditionally appending it - if you don't add syslogger-remote there the config loaded on line 81 won't work.  If you add it always it will get the iteration error I am trying to fix because the address on line 85 will be null

nasthagiri (approved):

PR #132, opened 2018-01-31
PLAT-1852 Switch to pytest, improve test coverage

jmbowman:
Switched from nose to pytest because nose wasn't correctly catching the initialization of models.py in coverage data (and because we're standardizing on pytest for assorted other reasons).  Also added an assortment of new tests that increase coverage from 65% to 74%.

The Django dependency will need to be broken out of requirements.txt for the tox configuration to work as intended, but that should probably be part of a larger requirements handing cleanup.  I'm just checking `tox.ini` in now because it's an appropriate place to store the pytest configuration.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,21 @@
+[run]
+data_file = .coverage
+source =
+    queue
+omit =
+    queue/management/commands/tests/*
+    queue/migrations/*
+    queue/tests/*
+
+[report]
+exclude_lines =
+   pragma: no cover
+   raise NotImplementedError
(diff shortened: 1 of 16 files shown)


bmedx (approved):
Seems good to my inexperienced eyes. Looks like there's a fair amount to do here eventually.

jibsheet (approved):
Yay, no more nose.
I may remove a handful of the management commands (they'll be unimportant once rabbit is removed) but I think you also covered a number of commands we'll be keeping.

PR #136, opened 2018-02-12
PLAT-1853 Support Django 1.11

jmbowman:
Support and test Django 1.9, 1.10, and 1.11:

* Fixed URL configuration to work with the deprecation removals in 1.10
* Switched management commands from optparse to argparse
* Fixed a query to use the actual field names instead of property aliases for them
* Used the correct `TestCase` subclass for some tests involving manual transaction commits
* Test each Django version in Travis

### DIFF
--- a/queue/lms_interface.py
+++ b/queue/lms_interface.py
@@ -78,8 +78,8 @@ def submit(request):
                                         queue_name=queue_name,
                                         xqueue_header=xqueue_header,
                                         xqueue_body=xqueue_body,
-                                        urls=urls_json,
-                                        keys=keys_json)
+                                        s3_urls=urls_json,
+                                        s3_keys=keys_json)
                 submission.save()
                 transaction.commit()  # Explicit commit to DB before inserting submission.id into queue
 
--- a/queue/management/commands/run_consumer.py
+++ b/queue/management/commands/run_consumer.py
@@ -63,10 +63,9 @@ def assign_queues_to_workers(num_workers, queue_assignments):
 
 
 class Command(BaseCommand):
-    args = "<worker count>"
     help = """
     Listens to all queues specified as being push-queues in the django
-    configuration with <worker count> processes
+    configuration
     """
 
     def handle(self, *args, **options):
@@ -92,7 +91,6 @@ class Command(BaseCommand):
             time.sleep(MONITOR_SLEEPTIME)
 
         log.info(' [*] All workers finished. Exiting')
-
 
     def monitor(self, workers):
         finished_workers = []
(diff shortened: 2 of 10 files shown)


queue/lms_interface.py
```diff
@@ -78,8 +78,8 @@ def submit(request):
                                         queue_name=queue_name,
                                         xqueue_header=xqueue_header,
                                         xqueue_body=xqueue_body,
-                                        urls=urls_json,
-                                        keys=keys_json)
+                                        s3_urls=urls_json,
+                                        s3_keys=keys_json)
```
doctoryes, line 82: How did this even work with the wrong names?

doctoryes (approved):
Easier than I thought it'd be!

queue/lms_interface.py
jmbowman: Maybe it didn't? I think it was just silently ignoring those keyword arguments until 1.10 started saying "sorry, those aren't actual field names you know".

jibsheet: It appears this was written as part of the Swift work.  You can see that one commit in https://github.com/edx/xqueue/pull/103 names it correctly in the model, but the field names are switched back later in the PR (but not the code that calls it).  The only "benefit" is that this code isn't in production, since the prod AMI builds from the release branch which is hugely behind the master branch.  https://github.com/edx/xqueue/tree/release  Not running master is a known issue, since when we tried to go to master, a bunch of external graders broke (probably related to this inability to pass in the download URLs honestly, although most entries in the submission table don't use that field)  details - https://openedx.atlassian.net/browse/OPS-1880

queue/management/commands/run_consumer.py
```diff
@@ -63,10 +63,9 @@ def assign_queues_to_workers(num_workers, queue_assignments):
 
 
 class Command(BaseCommand):
-    args = "<worker count>"
```
jibsheet, line 66: I realize this appears unused, but the default execution of this to start the second part of the xqueue app passed in a worker count
https://github.com/edx/configuration/blob/master/playbooks/roles/xqueue/templates/xqueue_consumer.conf.j2#L3
I think the right fix is just to pull that out of the supervisor runner for xqueue, but I'm trying to make sure it's really unused

jmbowman: It was probably used in previous versions of the code, but was silently ignored by the code in master before this PR.  The entire `assign_workers_to_queues` and `assign_queues_to_workers` functions seem to have been similarly orphaned.  Those should be deleted if we don't have immediate plans to use them again, but that was a little out of scope for this particular PR.

jibsheet: Agreed that we can probably delete more of it - I'd been trying to figure out how it was being used in production, but hadn't chased through it all yet.  I only noticed because the consumer stopped starting because it passed a count that now throws an error.  I fixed that on devstack but suspect it needs a broader set of deletes.

PR #142, opened 2018-02-16
Switch to Makefile and codecov (from rakefile and coveralls)

jibsheet:
- We haven't tested on Jenkins for a long time
  We use TravisCI (see .travis.yml)
- junit was a Jenkins plugin
- Remove rake references
- No longer have a rakefile
- Coveralls was configured, but never worked.  Just switch to codecov.


tox.ini
```diff
@@ -21,9 +21,12 @@ deps =
 passenv =
     TRAVIS*
 whitelist_externals =
-    rake
+    make
 commands =
-    {posargs:rake test}
+    py.test --cov --cov-report term-missing --ds=xqueue.test_settings queue
```
jibsheet: @jmbowman  This used to be different in the rakefile
  python -Wd -m pytest --cov --cov-report term-missing --ds=xqueue.test_settings --junitxml=#{XUNIT_FILE} queue

This version passes tests - is it ok?  I tried to get closer to what we do in the cookiecutter before I start solving the docker issue

tox.ini
```diff
@@ -20,10 +20,9 @@ deps =
     -r{toxinidir}/requirements/test.txt
 passenv =
     TRAVIS*
-whitelist_externals =
-    rake
 commands =
-    {posargs:rake test}
+    py.test --cov --cov-report term-missing --ds=xqueue.test_settings queue
```
jibsheet: Sorry @jmbowman I think I sent this once and a rebase removed it.
Objections to switching away from
  sh("python -Wd -m pytest --cov --cov-report term-missing --ds=xqueue.test_settings --junitxml=#{XUNIT_FILE} queue")
junit was jenkinsy, and the rest seems to build and pass tests

jmbowman: * Jenkins uses the generated XML file for an assortment of reporting, but I guess Travis only cares about "did the process exit without yielding an error code?"  So yeah, I guess we don't really need it here.
* I'm a little hesitant to remove the "python -Wd -m" prefix, because that catches warnings triggered while pytest is initializing.  In the previous runs here it was only catching a couple of deprecations warnings internal to pytest, but in edx-platform we've seen Django startup deprecation warnings hidden without that.  Once pytest is ready to start running test cases, it catches and shows us any appropriate warnings from then on by default.

jibsheet: Yes, the xml/html stuff appears to mostly be for codecov to consume and we don't need the old Jenkins testing files anymore (we haven't used it in ~4 years apparently).

Keeping the warnings, especially on older codebases like this makes a lost of sense, I've switched back to it.  I had been reading some of the cookiecutter and other libraries to try and figure out what's "modern" while I switch this around.

jibsheet:
Walked through the deletions with @feanil who gave an in person thumb with deleting a lot of ancient unused scripts/configs

PR #152, opened 2018-02-21
Run tests inside a docker container (local and on travis)

jibsheet:
Swap around make and tox

This allows make to run you just the current virtualenv again while
allowing things like travis to run tox across all the virtualenvs.

This also means we don't need the long test command in two places that
might drift later.

Improve (or remove) the various docs to encourage use of Docker Devstack


jibsheet:
FYI @jmbowman I reverted back to the makefile/tox you originally did (where make test is current virtualenv, and make test-all or tox directly will test across all the django versions).  Let me know if anything seems odd to you, but happily this appears to be running across all the django versions (although I'll reconfirm after this test run) using the docker container to get us 16.04.  It seems making the other IDAs test against various Django versions will be straightforward which is nice.

FYI @estute since you're enjoying watching all the old cruft go away

coryleeio (approved):

PR #153, opened 2018-02-22
Stop relying on old config variables

jibsheet:
Stop using SERVICE_VARIANT.  Finish switching over to the UPLOAD variables since S3_BUCKET and S3_PATH_PREFIX were deprecated prior to ginkgo.
https://github.com/edx/configuration/pull/4351


jibsheet:
Tested a sandbox build which generated files and started xqueue properly.
https://github.com/edx/configuration/pull/4351 is the already reviewed, but if you're interested, connected PR.

feanil (approved):
Are you planning on removing it from the log files as well?: https://github.com/edx/xqueue/blob/master/xqueue/logsettings.py

jibsheet:
Great question - for some reason ag didn't see it.

Do we care that we're changing the log line being fed into splunk?  I don't think I care, but mostly just checking.

PR #155, opened 2018-03-01
Add configurable code that talks to the database, not RabbitMQ

jibsheet:
There are some extra explanations in commit messages, but the main changes are:

* Add a temporary WABBITS setting that defaults to true (keep using RabbitMQ)
* Change consumer workers to have two Worker classes that split the connections and are instantiated based on WABBITS using a shim class
* Run tests twice (WABBITS true and then false, combining coverage to ensure we test both splits) - this was less copy and paste than a DDT/django.test.override_settings change for something that is purely temporary
* Filter the returned submissions based on how recently a submission was pulled or pushed to help avoid duplications (the rabbit code only has one chance pulling an item out of the rabbitmq queue before it stops being processes)
* We will want actual indexes on this table before going live for this, as some of our larger queues in prod can take a minute to query (since it basically does a large tablescan) but we'll want to shrink the tables before indexing them.


jibsheet:
FYI @estute  - This is one of the larger PRs that you might enjoy (although you'll probably enjoy the pending "delete most of this code" PR more)

queue/consumer.py
```diff
@@ -55,6 +58,15 @@ def get_single_unretired_submission(queue_name):
                     If no unretired item in the queue, return False
         submission: A single submission from the queue, guaranteed to be unretired
     '''
+
+    if settings.WABBITS:
```
coryleeio, line 62: nice

coryleeio (approved):

PR #172, opened 2018-03-22
Remove RabbitMQ

jibsheet:


jibsheet:
FYI @doctoryes, @estute because you've either been interested in this code or been harmed by it

queue/lms_interface.py
```diff
@@ -80,11 +80,7 @@ def submit(request):
                 submission.save()
                 transaction.commit()  # Explicit commit to DB before inserting submission.id into queue
 
-                if settings.WABBITS: # Continue to put this entry in RabbitMQ
```
doctoryes, line 83: Wascally!

doctoryes (commented):
Didn't do an extensive review. But generally seems great using deleted code/reqs as a measure.
It's not clear what's replacing RabbitMQ as a queueing mechanism - perhaps add some text about that somewhere in the XQueue README-type info?

jibsheet:
@doctoryes That's a great question - I have a ticket for "fully go through all the docs", but I'll add something to the README.  The gist is that we now just query the database.

queue/management/commands/queue_ungraded_submissions.py
```diff
@@ -115,3 +114,11 @@ def requeue_submissions(self, submission_qset):
                 submission.save()
             push_to_queue(submission.queue_name, str(submission.id))
         self.stdout.write("Queueing finished")
+
+def push_to_queue(queue_name, submission_id):
+    """
+    This used to use a consumer push_to_queue function and send data to RabbitMQ
+    Now that we don't use RabbitMQ, this is less useful, but it's heavily leveraged
+    in the tests, so stub this to test things until it can be cleaned up.
```
feanil, line 122: Is there a ticket for this cleanup? Can we link it here?

xqueue/test_settings.py
```diff
@@ -83,6 +74,4 @@
 ENV_ROOT = ROOT_PATH
 AUTH_FILENAME = 'test_auth.json'
 
-# Default keep testing with RabbitMQ
-# We rerun the tests in `make test` and travis to test with only a database
-WABBITS = True
+CONSUMER_DELAY = 1
```
feanil: Add a comment about the unit of this in the test_settings as well.

feanil (approved):

queue/management/commands/queue_ungraded_submissions.py
jibsheet: Yep - I wrote it up after getting this passing (there were a handful of things that weren't strictly needed to ship this but should get cleaned up)
I've added it to the comment block.

PR #175, opened 2018-03-26
Bump to Django 1.11

jibsheet:
Tests are passing
Manual testing on a sandbox of push/pull graders is working
Ready to review and test further on stage and prod

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,6 @@ sudo: required
 matrix:
   include:
     - python: 2.7
-      env: TOXENV=py27-django18
-    - python: 2.7
-      env: TOXENV=py27-django19
-    - python: 2.7
-      env: TOXENV=py27-django110
-    - python: 2.7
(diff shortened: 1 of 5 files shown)


jibsheet:
@jmbowman @doctoryes  I should be doing the 1.11 upgrade here late in the week, but query about tox.ini.  Since this runs tests on a docker container which has Django 1.8 I figure keeping the override/reinstall in tox.ini makes sense for now, followed by a later commit after the docker image updates that either uses one travis executor or still splits to two?  (tests + quality)

I'll need to test some things, but working on deploying the rabbit removal early this week.

jmbowman:
@jibsheet Keeping the tests and quality checks in separate Travis executors sounds good.  Pretty soon we'll want to add a Python 3 tox environment, and then Django 2.0.x.

coryleeio (approved):

doctoryes (approved):
Totally :shipit:

PR #183, opened 2018-04-03
Pin pip version to 9.0.3

xitij2000:
For Ginkgo deployments, we have seen a couple of sandboxes fail because they ended up with [pip 10 beta](https://github.com/pypa/pip/releases/tag/10.0.0b1) installed. This breaks the [installation of pika from git|https://github.com/edx/xqueue/blob/3b2469c881a76cf5f095def21f87eafe005b5653/requirements.txt#L14], possibly due to a bug in the beta release. With this PR we hope to pin the version of pip to 9.0.3 (the latest 9.x release) similar to how it's pinned for [edx-platform](https://github.com/edx/edx-platform/blob/4841ff7943d397b409b6eae21b4085342e23af08/requirements/edx/pre.txt#L10). 

**JIRA tickets**: OSPR-2330

**Sandbox URL**: NA

**Merge deadline**: None

**Testing instructions**:

* Deploy sandbox based on this branch. 
* The xqueue virtualenv should have pip 9 instead of pip 10

**Reviewers**
- [ ] @rocioar
- [ ] edX reviewer[s] TBD

### DIFF
--- a/pre-requirements.txt
+++ b/pre-requirements.txt
@@ -1,1 +1,2 @@
 setuptools==15.2
+pip==9.0.3


openedx-webhooks:
Thanks for the pull request, @xitij2000! I've created [OSPR-2330](https://openedx.atlassian.net/browse/OSPR-2330) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

If you like, you can add yourself to the [AUTHORS](https://github.com/edx/xqueue/blob/master/AUTHORS)      file for this repo, though that isn't required.     Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.

rocioar (approved):
👍

nedbat:
@jmbowman Thoughts?  This is a Ginkgo pull request.

xitij2000:
For context, the version of pip on master sandboxes is also upgrading to pip 10 beta. However since xqueue:master doesn't include the pika dependency installed from git those sandboxes aren't failing. I honestly don't know why pip 10 is being installed at all, given that it's in beta. Any insight on this would be appreciated!

jmbowman:
@xitij2000 Do you have a log for a failing installation?  My first suspicion is that something (maybe the default Ansible settings) is configuring pip to be ok with auto-upgrading packages and installing pre-release packages if present.  But hunting for that would be easier with the full log of a case where it happened.

xitij2000:
@jmbowman That was my thought as well, but I couldn't reproduce it locally with the same version of ansible running a similar set of steps. I haven't explored the logs yet, but I think all our recent sandboxes with master should have pip 10, so I have places to look. 

The error while installing pika from git using pip 10 is as follows:
```
Obtaining pika from git+https://github.com/pika/pika.git@a731347fc0#egg=pika (from -r /edx/app/xqueue/xqueue/requirements.txt (line 14))
--
Cloning https://github.com/pika/pika.git (to revision a731347fc0) to /edx/app/xqueue/venvs/xqueue/src/pika
fatal: Couldn't find remote ref a731347fc0
:stderr:   Did not find branch or tag 'a731347fc0', assuming revision or ref.
Command \"git fetch -q https://github.com/pika/pika.git a731347fc0\" failed with error code 128 in /edx/app/xqueue/venvs/xqueue/src/pika
```

jmbowman:
The problem installing pika looks like [this pip issue](https://github.com/pypa/pip/issues/5140), using the full commit hash might help with that.  But I do want to make sure that we're properly installing the desired versions of setuptools and pip.  I recently made some changes to the configuration repo to do that more consistently across all the services, but would like to make sure we're not accidentally doing something afterwards to install different versions.

xitij2000:
I had a closer look at one of the Ginkgo deployments, and at the state at which it failed the only (Python-based) services installed are ``edxapp``, ``notifier`` and ``xqueue`` (partial). Since edxapp pins the version of pip, the version installed in its virtualenv is pip 9.0.1 while the version of pip for both ``notifier`` and ``xqueue`` is 10.0.0b1 (the latest at the time of deployment).

jmbowman:
I added some notes on what's happening here to [this virtualenv issue](https://github.com/pypa/virtualenv/issues/1157).  This is fixed in the master branch of the configuration repo by pinning exact versions of virtualenv, setuptools, and pip for everything, so it shouldn't be a problem for Hawthorn.  Let us know if it crops up anywhere else for Ginkgo that urgently needs fixing.

xitij2000:
@jmbowman Thanks for your help!

PR #192, opened 2018-04-12
When running background tasks, you need to tag transactions

jibsheet:
If you don't tag anything, NR doesn't know what to instrument.
One key is not marking the while loop or it'll never return data.
I suspect when the ticket to refactor some of this code lands, we'll
have one or two more transactions
https://docs.newrelic.com/docs/agents/python-agent/supported-features/python-background-tasks

### DIFF
--- a/queue/consumer.py
+++ b/queue/consumer.py
@@ -14,6 +14,12 @@ from django.db.models import Q
 from django.utils import timezone
 from requests.exceptions import ConnectionError, Timeout
 
+try:
+    import newrelic.agent
+except ImportError:  # pragma: no cover
+    newrelic = None  # pylint: disable=invalid-name
+
+
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


queue/consumer.py
```diff
@@ -14,6 +14,12 @@
 from django.utils import timezone
 from requests.exceptions import ConnectionError, Timeout
 
+try:
+    import newrelic.agent
```
feanil, line 18: Newrelic is in the requirements, why do you need this try-catch?

queue/consumer.py
```diff
@@ -132,7 +138,11 @@ def run(self):
             push_time_filter = Q(push_time__lte=(datetime.now(pytz.utc) - timedelta(minutes=settings.SUBMISSION_PROCESSING_DELAY))) | Q(push_time__isnull=True)
             submission = Submission.objects.filter(push_time_filter, queue_name=self.queue_name, retired=False).order_by('arrival_time').first()
             if submission:
-                self._deliver_submission(submission)
+                if newrelic:
+                    task = newrelic.agent.BackgroundTaskWrapper(self._deliver_submission)
```
feanil: Why not just use the decorator to decorate the `_deliver_submission` function instead.

I'm looking at the example here: https://docs.newrelic.com/docs/agents/python-agent/supported-features/python-background-tasks

queue/consumer.py
jibsheet: There was some push towards letting folks not install newrelic if they want to punt it (I have no context other than all the edx-platform newrelic code does this now) so I kept it alive in the xqueue thing to avoid doing it later.
I don't know if this is still a push or not, but mostly it fell into "not too complicated"

queue/consumer.py
jibsheet: Yep, that was my original solution - see the note above about checking for newrelic installed apparently being something done in edx-platform and one other IDA (ecommerce?) for code that calls newrelic.agent methods.  I lack context on why we did it, but I just continued in that process since it was easy to do.

queue/consumer.py
feanil: I'm not aware of any such push but ok,

queue/consumer.py
feanil: Even if we don't use the decorator because of the conditional, any reason to not move the wrapper instantiation to outside the while loop?  Just make one task wrapper that you can call multiple times?

queue/consumer.py
jibsheet: I dragged out the PR, something licensing related https://github.com/edx/edx-platform/pull/14633

queue/consumer.py
jibsheet: Oh, yeah, that's a good point, I'll move it up and check that nothing odd happens on my sandbox.

jibsheet: Moved the task wrapper around and confirmed it still submits stats periodically for each call.

coryleeio (approved):

feanil (approved):

PR #205, opened 2018-04-19
This command was for putting things back in Rabbit

jibsheet:
I misread some of the code earlier, but while trying to refactor some of the tests, it seemed clear to me now that this is really just for taking failed requests and putting them back on Rabbit.  Since
failed submissions will still be available for graders up until MAX_NUMBER_OF_FAILURES before we auto-retire them, I don't see as much used for this command.


jibsheet:
@gsidebo  I believe you were the original contributor.  Do you think we still need it without Rabbit being in play?

jdmulloy (approved):

jdmulloy:
:fire:

gsidebo:
@jibsheet @jdmulloy The issue being addressed by this command was not Rabbit-related. We were experiencing errors where submissions were being dequeued from xqueue by [xqueue-watcher](https://github.com/edx/xqueue-watcher), and failures were occurring before the submission could be successfully graded or a success message could be processed by xqueue. Most of the failures had to do with graders in xqueue-watcher failing under load when we ran jobs to grade python coding submissions for a large class. xqueue is not designed to care about a submission after it's dequeued (unless something has changed since October). As a result, there would be a lot of manual labor involved in figuring out which submissions had failed and re-queueing them to attempt grading again. That's what this command was designed to do in an automated way. Our high-load grading jobs only occur in the fall, so we haven't needed this command since around October when this was merged, but we could end up needing it again. If it has been really burdensome to fix these tests (would help to know details), then I guess I would say it's OK to remove the command, but the underlying need for this command is not solved by Rabbit no longer being in play.

jibsheet:
@gsidebo They were dequeued from xqueue because they were removed from RabbitMQ as soon as they were pulled by xqwatcher ( XQueue didn't use any of the rabbitmq features that allow for things to requeue).  This is no longer true.  Submissions become available again 1 minute after a pull without a result (configurable to be longer, edx for example allows 10 minutes before resurfacing the submission).  It sounds like removing rabbit has solved your issue because xqwatcher will just find it reappearing under get_submission instead of needing to add it back to RabbitMQ.

The tests were hard to rewrite because they mostly relied on the function that pushes to rabbit being run to test that things were triggered, and that function was removed (in fact, producer.py itself is gone at this point).  I'll plan on merging this later in the week unless I've misunderstood your explanation of the problem (which made sense to me under the old XQueue model).

gsidebo:
@jibsheet OK I misunderstood. Sounds like this is safe to remove. Thanks for the rundown! 👍

jibsheet:
@gsidebo These changes will all be in Hawthorne, so if you folks want to chat about your own xqueue/xqwatchers before the fall, let us know and we'll figure out some time.

PR #214, opened 2018-05-02
OPEN-2226 Improve adherence to OEP-18 guidelines

jmbowman:
We came up with better pip-tools usage guidelines over the course of finalizing OEP-18, use them here:

* Put the command we'd actually like developers to run in the re-generation instructions for each generated requirements file
* Make sure we're actually using the same versions of dependencies between different contexts (production, testing, development, etc.)
* Break out the pip-tools dependencies for use in the Makefile targets

Django and pytest get minor version bumps in the process.


requirements/dev.txt
```diff
@@ -2,42 +2,45 @@
...
-idna==2.6                 # via requests
+idna==2.6
 isort==4.3.4
 mock==2.0.0
-more-itertools==4.1.0     # via pytest
+more-itertools==4.1.0
 mysql-python==1.2.5
 newrelic==3.2.0.91
 path.py==11.0.1
-pbr==4.0.2                # via mock
```
jibsheet, line 26: Can I be sad that we lost where the dep came from?  I don't see us cleaning it up in our sed script, so maybe pip-tools changed?

jibsheet (approved):

requirements/dev.txt
jmbowman: This is because `dev.in` is just an aggregation of pip-compile output files.  The dependency source data is still captured in those, but pip-compile doesn't preserve it across the file inclusion; it just knows that we specified this exact version in the input parameters.

jibsheet: Got it.  That makes sense.  Glad I got a free Django upgrade in this PR, I was planning to do that shortly anyway.  Merge at your leisure and I'll rebase my other dep upgrades.

PR #215, opened 2018-05-03
Switch to a YAML config file

jibsheet:
The way our YAML config file usage works, we export settings directly from
YAML into the settings namespace, instead of our previous json system which
would do

FOO = ENV_TOKENS.get('FOO', FOO)

and pick up the default FOO from settings.py if nothing came in from the
config file.

This all happens implicitly now, and is closer to the way the IDA configs
work.  Most config is handled in settings.py or in the ansible play and
devstack.py and production.py are just for imports and *maybe* an
occasional deployment default.

Normalize aws_settings.py into production.py instead

While aws_settings will continue to work, it's better to just use
--settings xqueue.production instead.


jibsheet:
FYI @pomegranited @bradenmacdonald that this is also coming.  It's a small change for openstack, basically, set everything in yml and it'll show up.  There's a new config var that can be passed in from ansible along with all the things you're already doing.  It also has the two old defaults in ansible rather than in the config itself.
If you want to avoid reading it all, there's a single commit in this PR that touches openstack.

coryleeio (approved):

bradenmacdonald:
Thanks @jibsheet - we'll take a look early next week.

xqueue/production.py
```diff
@@ -0,0 +1,67 @@
...
+        raise ImproperlyConfigured(error_msg)
+
+
+# Keep track of the names of settings that represent dicts. Instead of overriding the values in settings.py,
+# the values read from disk should UPDATE the pre-configured dicts.
+DICT_UPDATE_KEYS = ()
+
+CONFIG_FILE = get_env_setting('XQUEUE_CFG')
+with codecs.open(CONFIG_FILE, encoding='utf-8') as f:
+    config_from_yaml = yaml.load(f)
```
feanil: Use `safe_load` instead.

feanil (approved):
One small change but otherwise :rocket:

xitij2000:
@jibsheet  Sorry for the delayed reply here. I've had a chance to test this in a sandbox as well, and it seems to be running fine with OpenStack settings and SWIFT with a minor setting adjustment. Thanks! :+1:

jibsheet:
@xitij2000 Thanks for the review - this has now been fully merged and is running for us

PR #227, opened 2018-05-16
Add retire_old_submissions command

jibsheet:
The old command was "retire failed submissions" so I renamed it

New command requires a queue and allows you to also only retire old submissions.

OPS-3057

### DIFF
--- a/queue/management/commands/retire_failed_submissions.py
+++ b/queue/management/commands/retire_failed_submissions.py
@@ -13,7 +13,7 @@ class Command(BaseCommand):
     help = """
            Retire submissions that have more than settings.MAX_NUMBER_OF_FAILURES failures.\
            Notify the LMS/student that the queue will no longer attempt to process the submission.\
-           Optional <queue_name>
+           Optional <queue_name> - no queue name provided means all queues will be processed.
            """
 
     def add_arguments(self, parser):
--- a/queue/management/commands/retire_old_submissions.py
+++ b/queue/management/commands/retire_old_submissions.py
@@ -1,0 +1,47 @@
+import logging
+from queue.consumer import post_failure_to_lms
+from queue.models import Submission
+
+from django.core.management.base import BaseCommand, CommandError
+from django.utils import dateparse, timezone
+
+log = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = """
+           Retire submissions for the requested <queue> that were submitted from the LMS before <date>
+           It will send a message to the LMS so that learners know to resubmit, but will still be marked
+           as retired regardless of success contacting the LMS.
+           """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'queue_name',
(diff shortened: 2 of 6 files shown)


queue/management/commands/retire_failed_submissions.py
```diff
@@ -0,0 +1,59 @@
...
+
+class Command(BaseCommand):
+    help = """
+           Retire submissions that have more than settings.MAX_NUMBER_OF_FAILURES failures.\
+           Notify the LMS/student that the queue will no longer attempt to process the submission.\
+           Optional <queue_name>
+           """
+
+    def add_arguments(self, parser):
+        parser.add_argument('queue_name', nargs='*')
```
feanil: Add a help message saying that if no 'queue_name's are provided, then we will retire submissions from all queues.

queue/management/commands/retire_submissions.py
```diff
@@ -1,59 +1,47 @@
...
-            default=False,
-            help='Force retire submissions',
-        )
+            'queue_name',
+            help="Which queue to retire submissions for")
+
+        parser.add_argument(
+            '--retire-before',
+            dest='retire_before',
+            help='Date + time to limit submissions up for retirement (arrival_time < YYYY-MM-DD HH:MM::SS)'
```
feanil: Also mention that this time should be in UTC.

queue/management/commands/retire_submissions.py
```diff
@@ -1,59 +1,47 @@
 import logging
```
feanil: Maybe rename this command to `retire_old_submissions.py` for clarity?

feanil (approved):
Couple of nits, but generally looks good.

queue/management/commands/retire_failed_submissions.py
jibsheet: This is one of "positional - need a queue" args handled by management commands, seemed honestly more likely to me that you'd be retiring a single queue and I didn't want to accidentally do them all.
It'll error if you don't pass in a queue_name

queue/management/commands/retire_submissions.py
jibsheet: Good point - at one point that was here, but was lost when I cleaned up what style of date you needed

queue/management/commands/retire_submissions.py
jibsheet: Sure

queue/management/commands/retire_failed_submissions.py
feanil: If that's the case why do you have the if statement below that will try to retire all submissions if no queue_name is passed?  Should that get cleaned up?

jibsheet: Oh, sorry, I misread the name of the file and the diff was too tight to make it obvious.
This is the old command.  It does treat queue_name as optional.  My newer one treats it as "please pass one or Django /argparse will throw an error".
I'll add something to the help statement above.

PR #235, opened 2018-05-26
Make dependency setup friendlier to Dependabot

greysteil:
I noticed you're using Dependabot on this repo, but that it (Dependabot) is not quite up to handling your setup. In particular, it gets upset that your `requirements.txt` file is generated from `requirements/base.in` rather than from a `requirements.in` file. Would you be happy to change that? PR attached if so.

I'll also change Dependabot's file updating logic so it doesn't update the headers your adding (like it has done in https://github.com/edx/xqueue/pull/234/files).


openedx-webhooks:
Thanks for the pull request, @greysteil! I've created [OSPR-2835](https://openedx.atlassian.net/browse/OSPR-2835) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

We can't start reviewing your pull request until you've submitted a     [signed contributor agreement](http://open.edx.org/sites/default/files/wysiwyg/individual-contributor-agreement.pdf)     or indicated your institutional affiliation. Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.

natabene:
@greysteil  Sorry for long delay, our bot had a bug and this was missed. Thank you for the contribution! When you have a chance, could you please send in a signed Contributor Agreement mentioned in the earlier comment? You only need to do this once and we can then review all your future code contributions. Thanks

greysteil:
Sure thing - attached

[redacted]

natabene:
@greysteil  Thanks, can you actually send it to legal@edx.org ? Also, maybe you want to remove this attachment from the PR, as it contains your private contact information, and this repo is publicly accessible.

greysteil:
Sure thing - done.

greysteil:
@natabene - I've signed the contributor agreement now and have just rebased this - hopefully it's ready to review! :octocat:

natabene:
@greysteil  Great, thanks! @edx/testeng Can you take a look at this?

jmbowman:
The logic behind the current requirements management scheme is documented in [OEP-18](https://open-edx-proposals.readthedocs.io/en/latest/oep-0018-bp-python-dependencies.html) (OEP stands for "Open edX Proposal"); I think the Dependabot integration predates that, and unfortunately we've been largely ignoring those PRs recently.

What we now want from a dependency management automation tool is [summarized in that doc](https://open-edx-proposals.readthedocs.io/en/latest/oep-0018-bp-python-dependencies.html#automate-updates-of-exact-dependency-specifications).  I don't recall Dependabot working well with pip-compile the last time I looked, so it doing so now resolves one of the reasons we haven't been using it more.   It still doesn't seem to support multiple requirements files in the same repository though, which I think is the limitation that prompted this PR.  What we'd really like in a tool like this is the ability to check multiple requirements files for updates, and call our preferred command for updating all of them (`make upgrade` in this case) if a PR is needed due to dependency updates.  None of the tools I've seen support this in its entirety, so we've been planning to write our own scheduled job to do this, with the option of being triggered by a webhook sent by a service like Dependabot or requires.io instead.

Regarding the PR as written - it references a requirements.in file that doesn't exist yet, and wouldn't let Dependabot track any of the other requirements files.  Renaming requirements/base.* to requirements.* would let Dependabot track just the core dependencies, but wouldn't help us keep up with new versions of packages used for testing, development, etc.

greysteil:
Sorry, I should have provided more context! I've had a read through OEP-18.

Dependabot is capable of tracking multiple requirement files, and updating all of them in a single PR. That's what is happening already in PRs like [this one](https://github.com/edx/xqueue/pull/312/files). The logic there is that it picks up all the files in a `requirements` folder, and any files at the top level that end with `.txt` or `.in`.

Dependabot can't read your Makefile, so doesn't understand [these lines](https://github.com/edx/xqueue/blob/master/Makefile#L36-L47), but it can *nearly* perfectly figure out what order it needs to run `pip-compile` in to update your files successfully. The logic it uses to do so is [here](https://github.com/dependabot/dependabot-core/blob/master/lib/dependabot/update_checkers/python/pip/pip_compile_version_resolver.rb#L283-L345) - basically it figures out which files require each other and starts at the bottom of the tree working up.

What Dependabot *can't* figure our right now is that the `requirements.txt` file is generated by `requirements/base.in`, because it currently relies on the files having the same name (i.e., it needs it to be called `requirements/base.txt`. If it had that correct then it would update everything perfectly.

I can do some work on my (Dependabot's) side to fix that, but it's going to be quite hard to get the logic right. Are you saying that you're committed to having a `requirements.txt` at the top level and having it generated from `requirements/base.in`, and wouldn't consider moving either `requirements.txt` -> `requirements/base.txt` or `requirements/base.in` -> `requirements.in`?

greysteil:
Another alternative would be to use the save setup as [warehouse](https://github.com/pypa/warehouse/blob/master/requirements.txt) and edit `requirements.txt` to be just:

```
-r requirements/base.txt
```

Would that work (obviously with an edit to the Makefile, and a requirements/base.txt committed)?

jmbowman:
Ok, it sounds like there's a good solution here then.  `requirements.txt` was only being kept for backwards compatibility with existing scripts which expected to find it there, it really should be just an alias for `requirements/base.txt` as outlined in your last comment.  So I think the changes needed are:

* The `upgrade` target should generate `requirements/base.txt` instead of `requirements.txt`
* The reference to `requirements.txt` in `requirements/test.in` should be changed to `requirements/base.txt`
* `requirements.txt` should only contain an inclusion of `requirements/base.txt`

The part that Dependabot may still have trouble with is running the pip-compile commands in the correct order; for example, if it just does them alphabetically it'll regenerate `dev.txt` before `test.txt` which it includes.  It'll also lose the `CUSTOM_COMPILE_COMMAND` setting and the post-processing which allows tox to specify the Django version by excluding it from the test requirements.  Could Dependabot parse the command used to re-generate the requirements file from the file header comment, since that should usually be correct?  The default value is what you're already doing, but it can be customized via `CUSTOM_COMPILE_COMMAND` like we've done here to override that behavior.

greysteil:
OK - I've made those changes and this should now be good to merge. However, as you say, Dependabot still won't work perfectly because of the post-processing for `tox`.

What's the thinking on using `tox` to set the Django version here? It looks like you're only considering one version in tests, so is it necessary? I'm also wondering whether an alternative would be for the `upgrade` command to work on multiple `base.in` files, each corresponding to a different Django version, and then to have `tox` pick the correct file. That way there would be no danger of sub-dependency problems if/when you use tox to test against multiple Django versions.

greysteil:
And thanks masses for talking all of the above through with me. Python isn't my main language, so apologies if any of my suggestions seem obviously wrong.

(I think getting Dependabot to read your Makefile and/or use the correct command, rather than derive it, would be hard, btw, because Dependabot doesn't clone the repo - it only pulls down the files it *thinks* are going to matter. It can already figure out the right order to compile files, however (based on which files require each other).)

jmbowman:
We previously had this repo configured to run tests against both Django 1.8 and 1.11 (while we were transitioning from the previous LTS release to the current one), and within a year we'll have it configured to run against both 1.11 and 2.2 (the next LTS release).  Even if we omit Django from the `*.in` files, other packages depend on it, so pip-compile finds the dependency and pins it to a specific version.

Having separate requirements files for each set of tox dependencies could work, although it can start to get unwieldy.  I've worked on some packages which support all recent Django versions and also vary on versions of another dependency (like `celery`); that can be a lot of extra requirements files for something that's otherwise just a few lines in `tox.ini`.  And pip-compile starts getting pretty slow as the number of dependencies increases, so running it a few more times would make updating dependencies in larger projects a little more frustrating.  I agree it's a reasonable solution for this repo right now, though.

That just leaves getting the correct command to run listed at the top of each file.  While Dependabot may be analyzing the files to run pip-compile on all of them in the correct order, someone updating dependencies manually is likely to get it wrong unless the file reminds them to use a command that does them all in the correct sequence.  Also, we do have other repositories where a post-processing script has to be run against all the generated requirements files in order to work around current bugs and limitations in pip-compile.  So we are going to have cases where we either can't use Dependabot, or would need it to be more flexible about how it updates the requirements (either cloning the repo to run the command as given, or just calling a webhook of our creation when it detects that a dependency needs updating).

natabene:
@greysteil  Please let us know once you have addressed feedback and it is ready to be looked at again. Thanks!

feanil:
@greysteil closing this since we haven't heard back from you in a while.  Feel free to re-open this PR when you're able to come back to it.

openedx-webhooks:
@greysteil Even though your pull request wasn’t merged, please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/xqueue&entry.867055334=https://github.com/edx/xqueue/pull/235&entry.1484655318=https://github.com/greysteil&entry.752974735=2018-05-26+18:02&entry.1917517419=2019-06-11+19:29&entry.2133058324=No) so we can improve your experience in the future.

PR #545, opened 2019-09-03
run make upgrade

zainab-amir:
Python3 upgrade is failing https://github.com/edx/xqueue/pull/544

It is because of this package
```
futures==3.2.0
```
Python 3 users should not attempt to install it, since the package is already included in the standard library and that's why I have added the constraint.txt file


zainab-amir:
@feanil @azarembok kindly review

feanil (approved):

PR #575, opened 2019-10-02
Rename djangoapp app

zainab-amir:
"queue" app name conflicts with a module of urllib3 which is installed as a dependency for requests package.

- I have changed the django app name by keeping the original db name using db_table field in model meta. 
- Updated references to old app name in settings, makefile, coverage file
- Updated imports
- Updated tests to mock correct app functionality

PROD-389


feanil (changes requested):
Generally looks good, I would suggest changing the name of the new package to 'submission_queue' instead of `submit_queue` but once that's done I think this should be good to go.

feanil (approved):

feanil:
@zainab-amir changes look good but let's hold off on merging until we have documented a full deployment plan.

coryleeio:
Migrations are faked so i've merged this. I have a runbook in the associated ops ticket

PR #578, opened 2019-10-07
Change tox file to run python3

zainab-amir:
Update tox file to run tests on python2 and python3


zainab-amir:
@feanil kindly review

tox.ini
```diff
@@ -31,3 +31,15 @@ whitelist_externals =
 commands =
     {toxinidir}/script/max_pep8_violations
     make selfcheck
+
+[testenv:py35-quality]
```
feanil, line 37: @zainab-amir I'm not familiar with what you're doing here.  Can you explain what's going on?

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,35}-django111,py{36,37}-django{111,20,21}
+envlist = py{27,35}-django111, py{27,35}-quality
```
feanil: Why are you removing the ability to test with newer versions of python or django? I think it makes sense to add a quality shard but you should still make it so that tox can still run on newer versions of python and django as we'll need to upgrade both of those in the near future.  Django 1.11 goes out of support in April and Python 3.5 will go out of support in September 2020.

feanil (commented):
Have some comments/questions on the tox.ini that I would love to learn more about.

tox.ini
zainab-amir: According to [tox documentation](https://tox.readthedocs.io/en/latest/config.html#complex-factor-conditions) we should be adding quality shard/command inside the `[testenv]` with constraints when to run which command:
```
commands =
    django111: make test
    quality: {toxinidir}/script/max_pep8_violations
```
Other option is to create a separate name environment for python 2 and python 3 separately. Named environment by default use the `basepython`. In order to run it explicitly for different versions of python we have to make multiple named environments, like in this case I created py35-quality and py27-quality. For reference see [here](https://stackoverflow.com/questions/51739680/how-to-get-tox-to-test-named-test-environments-with-different-python-versions/52468970#52468970)
This can also be confirmed by checking the config of these environments using `tox --showconfig`. I have truncated the output to just show the basepython in each case

```
[testenv:py35-quality]
basepython = python3.5

[testenv:py27-quality]
basepython = python2.7

[testenv:quality]
basepython = /edx/app/xqueue/venvs/xqueue/bin/python
```

zainab-amir: I am not sure but I think the reason why [testenv:quality] works for other edX repositories is because those do not run test inside a docker image. Travis activates a virtual environment and they use that as basepython which picks the correct python version from virtual env.
I have created a test PR for xqueue where I have used "quality" as TOXENV for different version of python given in travis.yml.
```yaml
  include:
    - python: 2.7
      env: TOXENV=quality
    - python: 3.5
      env: TOXENV=quality
```
Test PR : https://github.com/edx/xqueue/pull/624
When you open the build output you will see that it activates the correct virtual environment but runs the quality tests on python 2 and gives a deprecation warning for python 2. It means it's running quality for python 2 in both cases

```shell
Setting environment variables from .travis.yml
$ export TOXENV=quality
0.01s$ source ~/virtualenv/python3.5/bin/activate
$ python --version
Python 3.5.6
$ pip --version
pip 18.1 from /home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/pip (python 3.5)
before_install.1
$ docker-compose -f .travis/docker-compose-travis.yml up -d
before_install.2
5.45s$ docker exec xqueue bash -c "source /edx/app/xqueue/venvs/xqueue/bin/activate; cd /edx/app/xqueue/xqueue/; pip install -r requirements/travis.txt"
install
21.29s$ pip install -r requirements.txt
5.34s$ docker exec xqueue bash -c "source /edx/app/xqueue/venvs/xqueue/bin/activate; cd /edx/app/xqueue/xqueue/; TOXENV=${TOXENV} tox"
quality create: /edx/app/xqueue/xqueue/.tox/quality
quality installdeps: -r/edx/app/xqueue/xqueue/requirements/quality.txt
quality installed: DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support,pycodestyle==2.5.0
```
As you can see it activates the correct python env but runs the tests on python 2 (the build for this PR quality tests doesn't show this warning)

feanil: I see is this because it runs the quality test under the docker container?

zainab-amir: pytests and quality both run under docker container

zainab-amir:
@feanil any update on this?

tox.ini
feanil: This question is still un-answered.

feanil (commented):
I'd like to add back python 3.6, and 3.7 testing for unit tests at-least since we'll be moving to that in the next year.

tox.ini
zainab-amir: I was just running tests for django 2.7 and 3.5 in travis so I removed the environments from tox.ini file as well. I have added these environments back to tox and will enable tests for them in Travis.

feanil (approved):

PR #611, opened 2019-11-12
Add replaces

zainab-amir:
Add replaces to avoid faking migrations


submission_queue/migrations/0002_stop_updating_arrival_time.py
```diff
@@ -7,6 +7,8 @@
 
 class Migration(migrations.Migration):
 
+    replaces = [('queue', '0002_stop_updating_arrival_tim')]
```
zainab-amir: `0002_stop_updating_arrival_time`

PR #659, opened 2020-01-13
Fixes failing of automatic make upgrade

adeelkhan:
Make upgrade is failing because of
a legacy wsgiref package not compliant
with python3. Removing it from xqueue
requirements

PROD-1168


requirements/constraints.txt
```diff
@@ -10,4 +10,4 @@
 
 # This packages is a backport which can only be installed on Python 2.7
 futures ; python_version == "2.7"
-wsgiref ; python_version == "2.7"
+wsgiref ; python_version == "2.7"
```
jmbowman: You should be able to remove this line now.

requirements/dev.txt
```diff
@@ -6,59 +6,60 @@
...
 path.py==11.0.1
 pathlib2==2.3.5
-pip-tools==4.2.0
-pluggy==0.13.0
-py==1.8.0
+pip-tools==4.3.0
+pluggy==0.13.1
+py==1.8.1
 pycodestyle==2.5.0
-pyinotify==0.9.6 ; "linux" in sys_platform
```
jmbowman, line 43: We want to keep this; you may need to manually add it back if you aren't using Linux locally, or just re-run the automated `make upgrade` job after merging (which should also restore it).

jmbowman (approved):
Two minor suggestions, otherwise 👍

requirements/constraints.txt
adeelkhan: Done

requirements/dev.txt
adeelkhan: I think what i understand from above is that as iam using mac machine to run make upgrade its removing pyinotify package.. Any person on linux machine would not be running make upgrade to regenerate it and his/her system would not get this package. I guess automatic make upgrade executes on linux machine and this would get regenerated then. I have added it again manually.

adeelkhan:
@jmbowman Two things i have done after your last review.
1) I have now run make upgrade within python 3 requirement, as automatic upgrade is using py 3 env.
2) I have added more-itertools==5.0.0 in constraints as there was an installation issue for py2. If you execute make upgrade on py3 it would generate more-itertools==8.1.0 which py2 env can't install (max value it supports is 5.0). Similarly for pytest. py3 is using pytest == 5.3.2 while py2 is on pytest=4.6.6. Some how i think we should remove py2 test from travis as they are no more applicable.

jmbowman (approved):
Yes, we should drop the Python 2 tests; that can be done later in a separate PR.

PR #676, opened 2020-01-24
BOM-1175

awais786:
Updating travis and tox.
BOM-1174 and 1175.
removed py2 support from tox and travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,31 +3,19 @@ services:
   - docker
 sudo: required
 
+python:
+- 3.5
+
+env:
+- TOXENV=django111
+- TOXENV=django20
+- TOXENV=django21
+- TOXENV=django22
+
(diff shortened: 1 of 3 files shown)


mraarif (approved):

.travis.yml
```diff
@@ -3,31 +3,30 @@ services:
   - docker
 sudo: required
 
+python:
+- 2.7
```
jmbowman: We should probably just drop Python 2.7 testing altogether; we've already upgraded this service in production.

jmbowman (approved):
One suggestion to cut down on the amount of work Travis needs to do, otherwise this looks good to go.  And given that tests are now passing with all Django 2.x versions, I'll talk to the service owner about scheduling an upgrade of this to Django 2.2 in production.

PR #703, opened 2020-02-18
BOM-1288

awais786:
Removed django11 from tox and travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,6 @@ python:
 - 3.5
 
 env:
-- TOXENV=django111
-- TOXENV=django20
-- TOXENV=django21
 - TOXENV=django22
 
 matrix:
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{35}-django{111,20,21,22}-quality
+envlist = py{35}-django{22}-quality
 skipsdist=True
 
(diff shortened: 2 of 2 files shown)


awais786:
There is no setup.py.

Ayub-Khan (approved):

.travis.yml
```diff
@@ -7,7 +7,6 @@ python:
 - 3.5
 
 env:
-- TOXENV=django111
 - TOXENV=django20
 - TOXENV=django21
```
jmbowman: We should remove the 2.0 and 2.1 tests also, since we're already on 2.2.

awais786: I think we should keep the tests as edx-platform is still on django1.11. So we accedently do not break any django functionality which is not backwords compateble ?

awais786: @jmbowman ^^

jmbowman: xqueue isn't a dependency of edx-platform, it's a separate service which is already running in production on Django 2.2.  So if there were going to be any inter-service communication issues based on the Django version in use, we should have seen that when we did the deployment, but it went smoothly.

waheedahmed:
Unit tests stopped running in travis builds after this change, fixed it in https://github.com/edx/xqueue/pull/723.

CC: @fysheets

PR #738, opened 2020-04-27
Make xqueue alerts say xqueue for clarity

jdmulloy:


jdmulloy:
When alerts come in they don't say xqueue, so the person getting the page has to know that the queue name is an xqueue queue and not a normal celery queue. This fixes that. It will require manually a one time clean up of the old alarms in cloudwatch.

azarembok (approved):

jdmulloy:
Thanks @azarembok

PR #770, opened 2020-07-31
Revert "Removed usages of package six"

stvstnfrd:
This reverts commit 375bea4eb9106de275c6022612eca3b4bfde3e03.

We suspect this is a cause of CR-2392.


fysheets (approved):

davidjoy (approved):

stvstnfrd:
And for the sake of being explicit, since I noticed this PR does _not_ re-add `six` to `requirements.txt`, I _did_ confirm that it was _not_ removed in the original PR and still exists.

PR #778, opened 2020-08-28
BOM-1997 : Use python 3.8 in docker

mzulqarnain1:
Relevant JIRA : https://openedx.atlassian.net/browse/BOM-1997

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,14 @@
 FROM ubuntu:xenial
 
 RUN apt update && \
-  apt install -qy git-core language-pack-en libmysqlclient-dev ntp libssl-dev python3.5 python3-pip python3.5-dev && \
-  pip3 install --upgrade pip setuptools && \
+  apt-get install -y software-properties-common && \
+  apt-add-repository -y ppa:deadsnakes/ppa && apt-get update && \
+  apt-get install git-core language-pack-en python3-pip libmysqlclient-dev ntp libssl-dev python3.8-dev python3.8-venv -qy && \
   rm -rf /var/lib/apt/lists/*
 
-RUN ln -s /usr/bin/pip3 /usr/bin/pip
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,12 +1,17 @@
...
 RUN apt update && \
-  apt install -qy git-core language-pack-en libmysqlclient-dev ntp libssl-dev python3.5 python3-pip python3.5-dev && \
-  pip3 install --upgrade pip setuptools && \
+  apt-get install -y software-properties-common && \
+  apt-add-repository -y ppa:deadsnakes/ppa && apt-get update && \
+  apt-get install -y curl && \
+  apt-get upgrade -qy && \
+  apt-get install git-core language-pack-en python3-pip libmysqlclient-dev ntp libssl-dev python3.8-dev python3.8-distutils -qy && \
+  curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && \
+  python3.8 get-pip.py && python3.8 -m pip install --upgrade pip setuptools && \
```
jmbowman: Was this copied from an implementation elsewhere?  Before we copy it everywhere, I'm wondering if the `ensurepip` approach described in https://packaging.python.org/tutorials/installing-packages/#ensure-you-can-run-pip-from-the-command-line would work and be a bit simpler.

mzulqarnain1: I get `No module named 'pip'`, It seems Python versions coming from deadsnakes repo don't have any pip module packed in it. 
Check this for reference
https://github.com/deadsnakes/issues/issues/87#issuecomment-499905633

jmbowman: Ok; it looks like there are other ways to tackle this, but they aren't really any cleaner.  Ubuntu 16.04 packaging and pip installation just don't really play nicely.  I'm fine with this until we get onto 20.04, which should happen pretty soon.

jmbowman (approved):

PR #783, opened 2020-10-13
Update Travis MySQL Image to 5.7

adzuci:
Updated the devstack and Travis database settings to use MySQL 5.7 instead of 5.6.  This triggered problems in one test due to how `multiprocessing` interacts with Django database connections, as described in https://stackoverflow.com/questions/24406092/when-using-multiprocessing-to-access-mysql-it-always-throws-the-following-error .  The test was updated to close the parent process's connection before forking the sub-processes.

Also, nothing ever shut down the worker processes explicitly once they were started, so added that to the test grader stub.

Finally, coverage combine started failing with a "No data to combine" message, even though it doesn't seem like anything relevant has changed in this PR and it was previously working on master. The command is redundant anyway since pytest-cov does the combining into a single .coverage file before pytest exits, so it was removed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ branches:
 before_install:
   - docker-compose -f .travis/docker-compose-travis.yml up -d
   - docker exec xqueue bash -c "cd /edx/app/xqueue/xqueue/; pip3 install -r requirements/travis.txt"
-
 install:
   - docker exec -t xqueue bash -c "
     sudo add-apt-repository ppa:deadsnakes/ppa -y;
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ matrix:
         TOXENV="py35-django22"
       after_success:
         - pip3 install -r requirements/travis.txt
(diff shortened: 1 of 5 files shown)


jmbowman:
So closing the DB connection before forking the processes did the trick, but the import has to happen right before closing it.  If you import the connection at the top of the file and then try closing it within a function, you're likely to have a reference to a stale connection used during initialization which has already been closed (and closing it again just silently does nothing).

Also, I noticed that nothing ever shut down the worker processes explicitly once they were started, so I added that to the test grader stub; without that, my pytest run was hanging for a good while after pytest was finished doing actual work.

Finally, `coverage combine` started failing with a "No data to combine" message, even though it doesn't seem like anything relevant has changed in this PR and it was previously working on master.  The command is redundant anyway since pytest-cov does the combining into a single `.coverage` file before pytest exits, so I just removed it and verified that the coverage file is still created and has appropriate data.

Cc @awais786 @iamsobanjaved

awais786 (approved):

jmbowman:
@nedbat I suspect we want to cherrypick this for Koa, but it isn't critical.  It just gets xqueue to run and pass tests with the correct version of MySQL, and also use the MySQL 5.7 container in devstack.

PR #802, opened 2020-12-15
BOM-2090 : Post Python 3.8 Clean Up

mzulqarnain1:
Relevant JIRA : https://openedx.atlassian.net/browse/BOM-2090

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,6 @@ script:
   - docker exec xqueue bash -c "cd /edx/app/xqueue/xqueue/ && tox -e ${TOXENV}"
 
 after_success:
-  - python3.8 -m pip install -r requirements/travis.txt
-  - docker exec xqueue bash -c "cd /edx/app/xqueue/xqueue/; coverage xml"
-  - codecov
+  - test $TOXENV="django22" &&
+    docker exec xqueue bash -c "cd /edx/app/xqueue/xqueue/; coverage xml"
+    && codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ script:
   - docker exec xqueue bash -c "cd /edx/app/xqueue/xqueue/ && tox -e ${TOXENV}"
 
 after_success:
-  - test $TOXENV="django22" &&
-    docker exec xqueue bash -c "cd /edx/app/xqueue/xqueue/; coverage xml"
-    && codecov
+  - test $TOXENV = "django22" &&
+    python3.8 -m pip install -r requirements/travis.txt &&
+    docker exec xqueue bash -c "cd /edx/app/xqueue/xqueue/; coverage xml" &&
+    codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,9 @@
 language: python
+python: 3.8
+env:
+  - TOXENV=django22
+  - TOXENV=django30
+  - TOXENV=quality
 services:
   - docker
 sudo: required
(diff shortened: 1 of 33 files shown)


UsamaSadiq (approved):

.travis.yml
```diff
@@ -1,4 +1,9 @@
 language: python
+python: 3.8
+env:
+  - TOXENV=django22
+  - TOXENV=django30
```
mraarif, line 5: don't you think these should be named as `py38-django22` and `py38-django30` currently tox isn't running as expected inside the container; logs show`django30` environment actually installing Django 2.2

mzulqarnain1: Fixed that problem, this syntax wasn't the problem, the culprit was the `make test` command which was overriding django version when installing from `dev.txt` file which still has django==2.2. I have added a separate make target for `test_requirements`.

.travis.yml
```diff
@@ -10,34 +15,11 @@ branches:
...
-        - docker exec xqueue bash -c "cd /edx/app/xqueue/xqueue/; coverage xml"
-        - codecov
-    - python: 3.5
-      env:
-        TEST_NAME=quality-python
-        TOXENV="quality"
+after_success:
+  - python3.8 -m pip install -r requirements/travis.txt
+  - docker exec xqueue bash -c "cd /edx/app/xqueue/xqueue/; coverage xml"
+  - codecov
```
jmbowman: We don't want to do this for the quality job, do we?  I think that's why it was added to just the unit test ones.

requirements/dev.txt
```diff
@@ -21,41 +21,36 @@ edx-django-release-util==0.4.4  # via -r requirements/test.txt
...
 pyparsing==2.4.7          # via -r requirements/test.txt, -r requirements/travis.txt, packaging
 pytest-cov==2.10.1        # via -r requirements/test.txt
 pytest-django==4.1.0      # via -r requirements/test.txt
-pytest==6.1.2             # via -r requirements/test.txt, pytest-cov, pytest-django
+pytest==6.2.0             # via -r requirements/test.txt, pytest-cov, pytest-django
 python-dateutil==2.8.1    # via -r requirements/test.txt, botocore
 python-memcached==1.59    # via -r requirements/test.txt
 python-termstyle==0.1.10  # via -r requirements/test.txt
 pytz==2020.4              # via -r requirements/test.txt, django
-pywatchman==1.4.1 ; "linux" in sys_platform  # via -r requirements/dev.in
```
jmbowman: I'm guessing you ran `make upgrade` on a Mac or such?  We need to keep this dependency to avoid performance problems in devstack.

submission_queue/management/commands/tests/test_count_queued_submissions.py
```diff
@@ -1,21 +1,21 @@
+from io import StringIO
 from submission_queue.models import Submission
 
 from django.core.management import call_command
 from django.test import TestCase
-from django.utils.six import StringIO
-from mock import call, patch
+from unittest.mock import call, patch
```
jmbowman: Did you also intend to remove the mock dependency from the requirement files?  And if we ever run isort on these files, it'll probably complain that the new import belongs above with the other standard library imports.

.travis.yml
mzulqarnain1: I've added condition for the django22 environment here.

requirements/dev.txt
mzulqarnain1: Re added.

submission_queue/management/commands/tests/test_count_queued_submissions.py
mzulqarnain1: I've removed mock from dependencies and fixed the import order in the files.

jmbowman (approved):

PR #803, opened 2021-01-06
upgrade gunicorn

Ali-D-Akbar:
[PROD-333](https://openedx.atlassian.net/browse/PROD-333)

Upgrading gunicorn above 19.5.0 which resolved potential vulnerability in process_headers.


stvstnfrd (approved):
Not tested locally, but seems like simple version bumps.

PR #886, opened 2023-04-13
build: Remove yanked codecov package (and coverage upload)

timmc-edx:
Stop trying to install the deprecated (and now yanked) codecov package. Also remove the CI step that was trying to use it, since we'd prefer to unblock CI now and restore coverage later. https://github.com/openedx/xqueue/issues/887 tracks restoring coverage testing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,9 +29,3 @@ jobs:
       run: |
         docker exec xqueue bash -c "cd /edx/app/xqueue/xqueue/; pip3 install -r requirements/ci.txt"
         docker exec xqueue bash -c "cd /edx/app/xqueue/xqueue/ && DB_HOST=${{ matrix.db-version }} tox -e ${TOXENV}"
-
-    - name: Code Coverage
-      if: matrix.tox-env=='django32'
-      run: |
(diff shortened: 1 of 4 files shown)


robrap (dismissed):

timmc-edx:
Hmm, looks like it was actually still invoking codecov: https://github.com/openedx/xqueue/blob/b8ef63b43e49d98a5fd64090944ada5b7af0b094/.github/workflows/ci.yml#L37

Maybe we want to just remove that coverage step for now.

robrap:
Can it just do what other services are doing?

timmc-edx:
It's doing all of its tests inside a docker container for some reason. The other repos that use the coverage action are able to do so because they're not running tests that way.

robrap:
Maybe just drop coverage and add an issue for changing this to better match other repos?

timmc-edx:
Yeah, I think that's the best route. It looks like we should be able to generate a coverage report inside the container, copy the files out of the container, and then run the action -- but I'd like to figure that out a different day. :-)

timmc-edx:
Created tracking issue: https://github.com/openedx/xqueue/issues/887

timmc-edx:
@robrap Ready for re-review.

robrap (approved):

PR #894, opened 2023-07-06
chore: install pkg-config for PyMySQL compatibility

leangseu-edx:
https://github.com/edx/edx-arch-experiments/issues/349

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM ubuntu:focal as app
 # System requirements
 
 RUN apt-get update && \
-    apt-get upgrade -qy && apt-get install language-pack-en locales git \
+    apt-get upgrade -qy && DEBIAN_FRONTEND=noninteractive apt-get install language-pack-en locales git \
     python3.8-dev python3-virtualenv libmysqlclient-dev libssl-dev build-essential pkg-config wget unzip -qy && \
     rm -rf /var/lib/apt/lists/*
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM ubuntu:focal as app
 
 RUN apt-get update && \
     apt-get upgrade -qy && apt-get install language-pack-en locales git \
-    python3.8-dev python3-virtualenv libmysqlclient-dev libssl-dev build-essential wget unzip -qy && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM ubuntu:focal as app
 
 RUN apt-get update && \
     apt-get upgrade -qy && apt-get install language-pack-en locales git \
```
timmc-edx:
```suggestion
    apt-get upgrade -qy && DEBIAN_FRONTEND=noninteractive apt-get install language-pack-en locales git \
```
Otherwise the docker build will ask for input and hang. (See notes in https://github.com/edx/edx-arch-experiments/issues/349)

Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM ubuntu:focal as app
...
-    python3.8-dev python3-virtualenv libmysqlclient-dev libssl-dev build-essential wget unzip -qy && \
+    python3.8-dev python3-virtualenv libmysqlclient-dev libssl-dev build-essential pkg-config wget unzip -qy && \
```
timmc-edx, line 7: It might be worth adding a comment to explain what these dependencies are for.
```suggestion
    python3.8-dev python3-virtualenv \
    # mysqlclient build dependencies
    libmysqlclient-dev libssl-dev build-essential pkg-config \
    wget unzip -qy && \
```

Dockerfile
timmc-edx: Actually, passing `--no-install-recommends` to apt-get install might be even better, since tzdata was only a _recommended_ transitive dependency. If it works, it's a good idea, since it will shrink your image sizes. But there's some chance you're depending on a recommended package without realizing it; up to you whether the risk is worth it.

timmc-edx (approved):
That should work. Remaining suggestions are optional. :-)

Dockerfile
leangseu-edx: I had success building the image with `DEBIAN_FRONTEND=noninteractive`. The problem is, I am not sure how to test these image. I do not want to take change on thing I can't test.

mattcarter (approved):
approved

Dockerfile
timmc-edx: Yeah, `DEBIAN_FRONTEND=noninteractive` should be safe. If xqueue is not something you can readily test, it's probably not worth shrinking the image size.

PR #905, opened 2023-09-26
fix: Actually install pip.txt when installing requirements

timmc-edx:
We were creating pip.txt but not using it, which eventually leads to errors when pip-tools falls out of sync with the system pip.


rgraber (approved):
LGTM

PR #976, opened 2025-07-16
fix: Replace default secret key with placeholder

timmc-edx:
This makes it clear that the field is unused and should be overridden.


timmc-edx:
@ormsbee Would you be able to take a look at this?

timmc-edx:
@feanil Would you be able to take a look at this? Just some cleanup.

feanil (approved):

ormsbee:
Thanks guys. Sorry I missed this.

PR #977, opened 2025-09-24
feat: Upgrading storages with new dict Django52

hunzlahmalik:
Deprecate STATICFILES_STORAGE and DEFAULT_FILE_STORAGE, and use STORAGE to upgrade to Django 5.2

### DIFF
--- a/xqueue/production.py
+++ b/xqueue/production.py
@@ -19,6 +19,15 @@ def get_env_setting(setting):
         raise ImproperlyConfigured(error_msg)
 
 
+STORAGES = {
+    "default": {
+        "BACKEND": "django.core.files.storage.FileSystemStorage",
+    },
+    "staticfiles": {
+        "BACKEND": "django.contrib.staticfiles.storage.StaticFilesStorage",
+    },
+}
+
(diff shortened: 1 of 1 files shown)


awais786 (approved):

xqueue/production.py
```diff
@@ -35,8 +43,19 @@ def get_env_setting(setting):
...
+    static_backend = config_from_yaml.pop('STATICFILES_STORAGE', None)
+
     vars().update(config_from_yaml)
 
+    if default_backend:
+        STORAGES['default']['BACKEND'] = default_backend
+
+    if static_backend:
+        STORAGES['staticfiles']['BACKEND'] = static_backend
+
```
feanil, line 57: Move this above the `vars().update...` so that if people do define the new storages dict in their config it takes precedence over the old setting values.

feanil (changes requested):

feanil (approved):

feanil:
Looks like the failure we're seeing here is also happening on master.

feanil:
I've created https://github.com/openedx/xqueue/pull/980 to fix the upstream issue.  Once that merges, I supect this will pass and we can merge this as well.

